PR #2, opened 2015-04-11
Fix a validator and param unmarshaling bug

analogue:
This should be two separate pull requests but hey it is Friday..

- Implement a custom 'type' validator so that required parameters with a value of None are skipped 
- Fix unmarshaling of optional parameters with default values
- Add jsonschema as a dependency in setup.py


bravado_core/swagger20_validator.py
```diff
@@ -18,24 +18,46 @@ def ignore(_validator, *args):
...
+        :class:`jsonschema.validators.Draft4Validator`
+    :param types: validate types
+    :type types: string or list
+    :param instance: object instance value
+    :param schema: swagger spec for the object
+    :type schema: dict
+    """
+    if is_param_spec(schema) and instance is None:
+        return
+    else:
```
prat0318: Remove `else`.

bravado_core/schema.py
```diff
@@ -22,7 +22,7 @@ def get_default(schema_object_spec):
 
 
 def is_required(schema_object_spec):
-    return 'required' in schema_object_spec
+    return schema_object_spec.get('required', False)
```
prat0318: The method should now be `get_required` if we are now fetching the value of `required`.

PR #3, opened 2015-04-13
Add config toggles for request validation and swagger spec validation

analogue:
Pretty self explanatory. Adding to support toggles in pyramid-swagger.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -21,6 +21,9 @@ CONFIG_DEFAULTS = {
     # On the client side, validate outgoing requests
     # On the server side, validate incoming requests
     'validate_requests': True,
+
+    # Use swagger_spec_validator to validate the swagger spec
+    'validate_swagger_spec': True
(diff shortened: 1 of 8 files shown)


bravado_core/spec.py
```diff
@@ -115,6 +119,9 @@ def build_api_serving_url(spec_dict, origin_url, preferred_scheme=None):
         supported by the API.
     :return: base url which services api requests
     """
+    if origin_url is None:
+        return ''
```
prat0318: We should be defaulting to api spec_dict url if `origin_url` is `None`. It looks like [request['url']](https://github.com/Yelp/bravado-core/blob/16a579302081c1374aa0a2b61c830085c1b94f8e/bravado_core/operation.py#L113) will become a url of type `/foo/bar` if `origin_url` is `None`.

prat0318:
lgtm apart from the query regarding `origin_url`.

PR #5, opened 2015-04-13
Allow lookup of an operation given request method and path

analogue:
Server-side impls will use to easily map an incoming request to an operation in the swagger spec.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -59,6 +59,9 @@ class Spec(object):
 
         self.responses = None
 
+        # Built on-demand - see get_op_for_request(..)
+        self._request_to_op_map = None
+
     @classmethod
     def from_dict(cls, spec_dict, origin_url=None, http_client=None,
                   config=None):
@@ -89,6 +92,27 @@ class Spec(object):
     def build_responses(self):
         log.warn('TODO: implement Spec::build_responses()')
         return {}
+
+    def get_op_for_request(self, http_method, request_path):
+        """
+        Return the Swagger operation for the passed in request http method
+        and path. Makes it really easy for server-side implementations to map
+        incoming requests to the Swagger spec.
+
+        :type http_method: http method of the request
+        :type request_path: full request path. e.g. /foo/{bar}/baz/{some_id}
+        :returns: the matching operation or None if a match couldn't be found
+        :rtype: :class:`bravado_core.operation.Operation`
+        """
+        if self._request_to_op_map is None:
+            self._request_to_op_map = {}
+            for resource in self.resources.itervalues():
+                for op in resource.operations.itervalues():
+                    key = (op.http_method, op.path_name)
+                    self._request_to_op_map[key] = op
+
+        key = (http_method.lower(), request_path)
+        return self._request_to_op_map.get(key, None)
 
 
 def build_api_serving_url(spec_dict, origin_url=None, preferred_scheme=None):
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -93,25 +93,26 @@ class Spec(object):
         log.warn('TODO: implement Spec::build_responses()')
         return {}
 
-    def get_op_for_request(self, http_method, request_path):
+    def get_op_for_request(self, http_method, path_pattern):
         """
         Return the Swagger operation for the passed in request http method
-        and path. Makes it really easy for server-side implementations to map
-        incoming requests to the Swagger spec.
+        and path pattern. Makes it really easy for server-side implementations
+        to map incoming requests to the Swagger spec.
 
-        :type http_method: http method of the request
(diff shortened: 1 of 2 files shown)


bravado_core/spec.py
```diff
@@ -90,6 +93,27 @@ def build_responses(self):
...
+        :type http_method: http method of the request
+        :type request_path: full request path. e.g. /foo/{bar}/baz/{some_id}
+        :returns: the matching operation or None if a match couldn't be found
+        :rtype: :class:`bravado_core.operation.Operation`
+        """
+        if self._request_to_op_map is None:
+            self._request_to_op_map = {}
+            for resource in self.resources.itervalues():
+                for op in resource.operations.itervalues():
+                    key = (op.http_method, op.path_name)
```
prat0318: Should be `op.http_method.lower()` as well?

bravado_core/spec.py
```diff
@@ -90,6 +93,27 @@ def build_responses(self):
...
+                    self._request_to_op_map[key] = op
+
+        key = (http_method.lower(), request_path)
+        return self._request_to_op_map.get(key, None)
```
prat0318: I don't see why generating the complete dictionary and then matching up is of any advantage instead of returning back the `op` as soon as you find it.

prat0318: I think it should also take into account partial matches as done in `pyramid_swagger` : https://github.com/striglia/pyramid_swagger/blob/master/pyramid_swagger/model.py#L60

bravado_core/spec.py
analogue: According to the [spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#path-item-object) it is required to be lowercase.

bravado_core/spec.py
analogue: Can you give me an example of a partial match? Two types of matches that I can see:

A path with static path components
```
route_pattern='/foo/bar/baz'
request.path='/foo/bar/baz'
```

A path with dynamic path components
```
route_pattern='/foo/{id}/baz'
request.path='/foo/1/baz'
```

For reference, [route docs](http://docs.pylonsproject.org/docs/pyramid/en/latest/narr/urldispatch.html#route-configuration)

analogue: I do think that the function argument named `request_path` is not clear as it could be.  `request_pattern` would more clearly indicate that `{blah}` style dynamic path components are expected.

bravado_core/spec.py
```diff
@@ -90,6 +93,27 @@ def build_responses(self):
...
+        :type request_path: full request path. e.g. /foo/{bar}/baz/{some_id}
+        :returns: the matching operation or None if a match couldn't be found
+        :rtype: :class:`bravado_core.operation.Operation`
+        """
+        if self._request_to_op_map is None:
+            self._request_to_op_map = {}
+            for resource in self.resources.itervalues():
+                for op in resource.operations.itervalues():
+                    key = (op.http_method, op.path_name)
+                    self._request_to_op_map[key] = op
```
dnephin: Caching in the library itself seems like it's not really exposing the best possible interface. Could we change this to instead expose a `build_op_for_request()` (which is the bulk of this method), and let the consumer of this library handle whatever caching makes sense for it?

analogue: I actually wrote it that way initially but then the client has another object to keep track of -- it already has `bravado_core.spec.Spec`. The cache is never invalidated and is not meant to be invalidated, hence caching is not a concern on the client side. This is more lazy initialization than anything else.

bravado_core/spec.py
prat0318: But the place where it is been [called from](https://github.com/striglia/pyramid_swagger/pull/100/files#diff-2b77869570ba0feb25040fddd8b35816R185) would have a path like `/foo/1/baz`, where will it be coverted to `/foo/{id}/baz` for it to match with the swagger spec path?

analogue: Ok, I think I understand where the disconnect is.  So the place where it is called from  actually has a path like `/foo/{id}/baz`.  Notice, that data is extracted from the `route` info (this is part of pyramid's internal URLDispatch implementation), not `request.path` which contains `/foo/1/baz`. Make sense?

prat0318: Ah ok. One doubt still lingering is how pyramid-swagger is aware of `/foo/{id}/baz`. I do not see in striglia/pyramid_swagger#100 where are we registering all the swagger spec operations as `add_routes` so that `pyramid_swagger` will know that it has to convert a request of the form `/foo/1/baz` (which it receives over the wire) to the form `/foo/{id}/baz`.

analogue: That registration happens in the application that is using pyramid_swagger. I'll paste some links in IRC.

PR #6, opened 2015-04-15
Add API for validating outgoing responses

analogue:
`bravado_core.response.validate_response(...)`

Response transformation is "complicated"  - See SRV-1708. This is just validation.


bravado_core/response.py
```diff
@@ -99,3 +130,65 @@ def get_response_spec(status_code, op):
...
+
+    if response.content_type not in op.produces:
+        raise SwaggerMappingError(
+            "Response content-type '{0}' is not supported by the Swagger "
+            "specification's content-types '{1}"
+            .format(response.content_type, op.produces))
+
+    if response.content_type == APP_JSON:
+        response_value = response.json()
+        validate_schema_object(response_body_spec, response_value)
```
prat0318: Is the response been validated again in `pyramid_swagger`? And what about the `pyramid_swagger` flag `validate_response` was `False`?

bravado_core/response.py
```diff
@@ -99,3 +130,65 @@ def get_response_spec(status_code, op):
...
+    """
+    # response that returns nothing in the body
+    response_body_spec = response_spec.get('schema')
+    if response_body_spec is None:
+        if response.text in (None, ''):
+            return
+        raise SwaggerMappingError(
+            "Response body should be empty: {0}".format(response.text))
+
+    if response.content_type not in op.produces:
```
prat0318: This means that `produces` should never be empty in the swagger spec for it to succeed. It should be more like a warning instead of an error imo.

bravado_core/response.py
```diff
@@ -99,3 +130,65 @@ def get_response_spec(status_code, op):
             "for {1}. Either add a response specifiction for the status_code "
             "or use a `default` response.".format(op, status_code))
     return response_spec
+
+
+def validate_response(response_spec, op, response):
```
prat0318: I guess this will be called from outside, like from `pyramid_swagger`?

bravado_core/response.py
analogue: Ah, I forgot to check the config flag. Will fix.  Validation only happens once.

bravado_core/response.py
analogue: From my interpretation this means that `produces` can only be empty when the response body is `None` or empty.  L159 covers this case.  By the time L165 is reached, a non-empty body is present.  I don't see how it can be validated if the `Content-Type` is not specified, hence the error.

bravado_core/response.py
analogue: Correct

PR #8, opened 2015-04-20
Replace jsonref proxies

analogue:
Fixes compatibility problems with python 2.7 and other 3rd party json libraries that can't handle `JsonRef` objects.

### DIFF
--- a/tests/spec/replace_jsonref_proxies_test.py
+++ b/tests/spec/replace_jsonref_proxies_test.py
@@ -1,0 +1,62 @@
+import json
+import jsonref
+
+from bravado_core.spec import replace_jsonref_proxies
+
+
+def test_simple():
+    d = {
+        'foo': {
+            "$ref": "#/definitions/user"
+        },
+        'definitions': {
+            'user': {
+                'properties': {
+                    'first_name': {
+                        'type': 'string'
+                    }
+                }
+            }
+        }
+    }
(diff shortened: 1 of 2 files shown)


tests/spec/replace_jsonref_proxies_test.py
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                }
+            }
+        }
+    }
+    json_obj = jsonref.loads(json.dumps(d))
+    assert isinstance(json_obj['foo'], jsonref.JsonRef)
+
+    replace_jsonref_proxies(json_obj)
+    assert isinstance(json_obj['foo'], dict)
```
prat0318: How about also an assert equal with `d['definitions']['user']` to be sure of the final result of the conversion?

tests/spec/replace_jsonref_proxies_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_nested():
```
prat0318: I would also add another test for `is_list_like` flow as well.

tests/spec/replace_jsonref_proxies_test.py
analogue: ack

tests/spec/replace_jsonref_proxies_test.py
analogue: ack

PR #12, opened 2015-05-07
For backwards compatibility with swaggerpy 1.2, make sure future.result() behaves identically

analogue:
- Moved HttpFuture to bravado
- unmarshaling a response now returns just the return value from the swagger response schema
- See <TODO> PR on bravado for the rest of the functionality


dnephin:
LGTM. Does it need a version bump as well so we can dep this version in the other packages?

analogue:
I'm going to do a version bump as soon as I finish verifying the end-to-end-integration with bravado. Shouldn't be too far off now.

PR #13, opened 2015-05-11
Fix HTTP method - should always be uppercase

analogue:
http methods are lower case in the swagger schema but required to be uppercase on the wire according to the HTTP RFC - http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.1

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -127,7 +127,7 @@ class Operation(object):
         """
         request_options = kwargs.pop('_request_options', {})
         request = {
-            'method': self.http_method,
+            'method': self.http_method.upper(),
             'url': self.swagger_spec.api_url.rstrip('/') + self.path_name,
             'params': {},
             'headers': request_options.get('headers', {}),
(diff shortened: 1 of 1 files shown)


bravado_core/operation.py
```diff
@@ -127,7 +127,7 @@ def construct_request(self, **kwargs):
         """
         request_options = kwargs.pop('_request_options', {})
         request = {
-            'method': self.http_method,
+            'method': self.http_method.upper(),
```
prat0318: How about doing it during the assignment of `self.http_method = http_method` as i see another place it is being changed to upper() [here](https://github.com/Yelp/bravado-core/blob/16a579302081c1374aa0a2b61c830085c1b94f8e/bravado_core/docstring.py#L90)

analogue: It is unfortunate the Swagger 2.0 Spec mandates the http method be lowercase.  In this case, since the context is an Operation as built from a Swagger Schema, returning exactly what the schema states is preferable.  The other place you mention the http_status being uppercased is for the presentation as a docstring, hence reasonable.

dnephin:
lgtm

PR #14, opened 2015-05-12
Support py3

### LINKED ISSUE #10: Support for python 3.x
https://github.com/striglia/pyramid_swagger would benefit from 3.x support.

Partial progress made here: https://github.com/analogue/bravado-core/tree/py3x_support

bpicolo:
This is ready once Julian/jsonschema#203 is in a release. Until then there's a bit of test flakiness for py3.

Fixes #10

### DIFF
--- a/bravado_core/docstring.py
+++ b/bravado_core/docstring.py
@@ -1,5 +1,7 @@
 import logging
 
+from six import iteritems
+
 from bravado_core.exception import SwaggerMappingError
 
 
@@ -100,7 +102,7 @@ def create_operation_docstring(op):
     # TODO: add 'examples' if available
     # TODO: good idea to identify the default response?
     responses = op_spec.get('responses')
-    for http_status_code, response_spec in iter(sorted(responses.iteritems())):
+    for http_status_code, response_spec in iter(sorted(iteritems(responses))):
         response_desc = response_spec.get('description')
         s += ':returns: {0}: {1}\n'.format(http_status_code, response_desc)
         schema_spec = response_spec.get('schema')
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -1,3 +1,5 @@
+from six import iteritems
+
 from bravado_core import formatter, schema
 from bravado_core.exception import SwaggerMappingError
 from bravado_core.model import is_model, MODEL_MARKER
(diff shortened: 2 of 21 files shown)


bravado_core/docstring.py
```diff
@@ -100,7 +100,7 @@ def create_operation_docstring(op):
     # TODO: add 'examples' if available
     # TODO: good idea to identify the default response?
     responses = op_spec.get('responses')
-    for http_status_code, response_spec in iter(sorted(responses.iteritems())):
+    for http_status_code, response_spec in iter(sorted(responses.items())):
```
prat0318: We can use `six.iteritems` here : https://pythonhosted.org/six/#six.iteritems

bravado_core/marshal.py
```diff
@@ -111,7 +111,7 @@ def marshal_object(swagger_spec, object_spec, object_value):
             type(object_value), object_value))
 
     result = {}
-    for k, v in object_value.iteritems():
+    for k, v in object_value.items():
```
prat0318: Same as above.

bravado_core/operation.py
```diff
@@ -154,7 +154,7 @@ def construct_params(self, request, op_kwargs):
             marshal_param(param, param_value, request)
 
         # Check required params and non-required params with a 'default' value
-        for remaining_param in current_params.itervalues():
+        for remaining_param in current_params.values():
```
prat0318: `six.itervalues`

bravado_core/param.py
```diff
@@ -1,9 +1,14 @@
 from functools import partial
 import logging
-import urllib
+try:
```
prat0318: `from six.moves.urllib.parse import quote`

bravado_core/docstring.py
bpicolo: we can, but me and spatel talked and reckoned it's not worth it. Speed and memory difference at our scales more or less negligible

bravado_core/param.py
bpicolo: Why use six when the builtin is easy to use? : P

analogue:
@benp thanks for picking this up and making it happen. lgtm

dnephin:
lgtm (once the jsonschema dep is ready and tests are stable)

bpicolo:
Bumped it and added a test difference for py3 for a new test : ) Should be good now methinks

bravado_core/operation.py
prat0318: i think if we have changed all `iteritems` to be used from `six` package, we should also change the `values()` to be changed to the `six` package to keep the consistency.

prat0318:
lgtm apart from the `values` concern.

prat0318:
:+1:

Issue #19, opened 2015-05-22
Refactor exception hierarchy


analogue:
Currently, bravado-core raises three types of exceptions when errors are encountered:
- SwaggerMappingError
- TypeError
- Exception

When either TypeError or Exception is raised, it is hard for a user of the library to determine if the error is Swagger related or a general Python error.  Replace TypeError and Exception with bravado-core specific exceptions that conform to a hierarchy with SwaggerError at the root.

SwaggerError
- SwaggerMappingError
- SwaggerTypeError
- etc

PR #20, opened 2015-05-22
Refactor exception hierarchy

analogue:
Fix for #19 

Base exception class is SwaggerError.
- SwaggerMappingError thrown when processing requests/responses
- SwaggerSchemaError thrown when processing the schema

I'll have follow up PRs for bravado and pyramid-swagger based on the change in the exception interface.


jnb:
:+1:

PR #21, opened 2015-05-26
Use basePath when matching a request to an operation

analogue:
See https://github.com/striglia/pyramid_swagger/issues/115


dnephin:
lgtm

PR #23, opened 2015-05-28
Rename ResponseLike to IncomingResponse for consistency

analogue:
Completing simple TODO to rename this class.  Associated PR for bravado is https://github.com/Yelp/bravado/pull/149


dnephin:
LGTM, 

Technically a non-backwards compatible change to the public interface, but since we're still in rc, I think that's fine

Issue #17, opened 2015-05-19
Refactor HTTP invocation to bravado


dnephin:
Just surfacing a TODO from [bravado_core/operation.py#L165](https://github.com/Yelp/bravado-core/blob/f93f884e5b09761d4f82f73dda1e56806f858a31/bravado_core/operation.py#L165) as a ticket.

Pretty low priority

PR #31, opened 2015-07-09
Move http invocation on Operation to bravado

analogue:
Fix for issue #17.  Linked PR on bravado the side is https://github.com/Yelp/bravado/pull/159


bravado_core/resource.py
```diff
@@ -93,7 +92,7 @@ def __getattr__(self, item):
         if not op:
             raise AttributeError(u"Resource '%s' has no operation '%s'" %
                                  (self.name, item))
-        return operation_docstring_wrapper(op)
```
prat0318: Why no more docstring wrapper? If i am not wrong, this is the only place that method is invoked.

prat0318:
apart from the query ^, lgtm

bravado_core/resource.py
analogue: docstring wrapper has to be the last decorator in the chain since it returns a function and not an instance of `bravado_core.operation.Operation`.

dnephin:
looks good, `bravado_core/http_client.py` is not internally referenced either, can that be moved as well?

analogue:
Good catch. Done.

dnephin:
lgtm

PR #33, opened 2015-07-10
[Rough Draft] adding User defined format

### LINKED ISSUE #32: Add UserFormatterLike interface
This is to be called from `pyramid_swagger` during [init](https://github.com/striglia/pyramid_swagger/blob/master/pyramid_swagger/__init__.py#L13)

Roughly interface should look like:

```python
class UserFormatterLike(object):
    """
    Interface for marshalling/unmarshalling user defined formatter objects.
    Subclasses are responsible for providing attrs for __required_attrs__.
    """
    __required_attrs__ = [
        'format',         # string - format name
        'description', # string - describes the marshalling & unmarshalling format
    ]

    def __getattr__(self, name):
        """
        When an attempt to access a required attribute that doesn't exist
        is made, let the caller know that the type is non-compliant in its
        attempt to be `UserFormatterLike`. This is in place of the usual throwing
        of an AttributeError.
        Reminder: __getattr___ is only called when it has already been
                  determined that this object does not have the given attr.
        :raises: NotImplementedError when the subclass has not provided access
                to a required attribute.
        """
        if name in self.__required_attrs__:
            raise NotImplementedError(
                'This UserFormatterLike type {0} forgot to implement an attr '
                'for `{1}`'.format(type(self), name))
        raise AttributeError(
            "'{0}' object has no attribute '{1}'".format(type(self), name))

    def to_python(self, **kwargs):
        """
        :return: unmarshalled python object
        """
        raise NotImplementedError("Implement to_python() in {0}".format(type(self)))

    def to_wire(self, **kwargs):
        """
        :return: marshalled python object
        :rtype: str
        """
        raise NotImplementedError("Implement to_wire() in {0}".format(type(self)))
```

and a method similar to `register_format` defined [here](https://github.com/Yelp/bravado-core/blob/master/bravado_core/formatter.py#L12):
```python
def register_formatter(formatter):
    """Register a user defined formatter with bravado-core

    :param formatter: object having information to marshal/unmarshal particular format
    :type formatter: :class:`UserFormatterLike`
    """
    _formatters[formatter.format] = (formatter.to_wire, formatter.to_python, formatter.description)    
```

prat0318:
- Fixes #32

- From comments of #32, I did not see an advantage of extra `validator` callback method over the logic of `to_python`. Even if the user provides `validator`, bravado-core cannot constrain that the raised exception will always be `ValidationError` (and secondly where will that be stored `_formatters` structure had to be changed). I think it is better to explicitly wrap it up. The only advantage we are getting with wiring in `validate` here is that a particular exception (ValidationError) is guaranteed in case of failure.

Related changes in `pyramid_swagger` will be [something like these](https://github.com/prat0318/pyramid_swagger/compare/master...add_user_formatter).

Opinions - @analogue @dnephin

### DIFF
--- a/bravado_core/exception.py
+++ b/bravado_core/exception.py
@@ -1,3 +1,7 @@
+import sys
+
+import six
+
 class SwaggerError(Exception):
     """Base exception class which all bravado-core specific exceptions
     inherit from.
@@ -22,3 +26,16 @@ class SwaggerSchemaError(SwaggerError):
     """Raised when an error is encountered during processing of a SwaggerSchema.
     """
     pass
+
+def wrap_exception(exception_class):
+    def generic_exception(method):
+        def wrapper(*args, **kwargs):
+            try:
+                method(*args, **kwargs)
+            except Exception as e:
+                six.reraise(
+                    exception_class,
+                    exception_class(str(e)),
+                    sys.exc_info()[2])
+        return wrapper
+    return generic_exception
--- a/bravado_core/formatter.py
+++ b/bravado_core/formatter.py
@@ -3,6 +3,7 @@ Support for the 'format' key in the swagger spec as outlined in
 https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#dataTypeFormat
 """
 import dateutil.parser
+from collections import namedtuple
 
 import six
 
@@ -20,6 +21,13 @@ def register_format(format, to_wire, to_python, description=None):
     :param description: useful description
     """
     _formatters[format] = (to_wire, to_python, description)
+
+
+def get_formatter(format):
+    """
+    Returns the appriopriate formatting tuples or None
+    """
+    return _formatters.get(format)
 
 
 def to_wire(spec, value):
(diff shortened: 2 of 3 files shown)


analogue:
How would this work for a user defined format?

```json
"type": "string"
"format": "palindrome"
```
With values of ```moon``` and ```noon```.

prat0318:
Hmm... i see the validator logic can be different if `to_python` is a no-op like in this case. So, i think we then need to store the `validate` entry as the fourth tuple value in `_formatters`?

analogue:
I think this is an opportunity to refactor the implemention to be easier to consume by bravado and pyramid_swagger.

bravado-core would encapsulate the necessary requirements in a named tuple
```python
SwaggerFormat = collections.namedtuple('SwaggerFormat', 'format to_python to_wire validate description)
```
- refactor `_defaults` to use the namedtuple
- refactor `register_format` to accept `SwaggerFormat`
- clients and services would both pass in a list of `SwaggerFormat`s to bravado and pyramind_swagger respectively -- keyed as ```user_defined_formats```
- document expectations of each attribute in the named tuple

Sound reasonable?

prat0318:
Sounds good. For clients and services to create list of `SwaggerFormat` either they would need to import `bravado-core` OR we can add a import of `SwaggerFormat` to `bravado` and `pyramid_swagger` which can then be imported by client and services respectively. Second looks like the winner to me.

PR #34, opened 2015-07-15
Support and validation for user defined Swagger formats

### LINKED ISSUE #32: Add UserFormatterLike interface

prat0318:
- add `SwaggerFormat` namedtuple
- add call to `validate` during primitive check
- Fixes #32

### DIFF
--- a/bravado_core/exception.py
+++ b/bravado_core/exception.py
@@ -1,3 +1,8 @@
+import sys
+
+import six
+
+
 class SwaggerError(Exception):
     """Base exception class which all bravado-core specific exceptions
     inherit from.
@@ -22,3 +27,17 @@ class SwaggerSchemaError(SwaggerError):
     """Raised when an error is encountered during processing of a SwaggerSchema.
     """
     pass
+
+
+def wrap_exception(exception_class):
+    def generic_exception(method):
+        def wrapper(*args, **kwargs):
+            try:
+                method(*args, **kwargs)
+            except Exception as e:
+                six.reraise(
+                    exception_class,
+                    exception_class(str(e)),
+                    sys.exc_info()[2])
+        return wrapper
+    return generic_exception
--- a/bravado_core/exception.py
+++ b/bravado_core/exception.py
@@ -7,7 +7,6 @@ class SwaggerError(Exception):
     """Base exception class which all bravado-core specific exceptions
     inherit from.
     """
-    pass
 
 
 class SwaggerMappingError(SwaggerError):
@@ -15,21 +14,24 @@ class SwaggerMappingError(SwaggerError):
     a response.
     """
 
-    def __init__(self, msg, cause=None):
-        """
-        :param msg: String message for the error.
-        :param cause: Optional exception that caused this one.
-        """
-        super(Exception, self).__init__(msg, cause)
+
+class SwaggerValidationError(SwaggerMappingError):
+    """Raised when an error is encountered during validating user defined
+    format values in a request or a resposne.
+    """
 
 
 class SwaggerSchemaError(SwaggerError):
     """Raised when an error is encountered during processing of a SwaggerSchema.
     """
-    pass
 
 
 def wrap_exception(exception_class):
+    """Helper decorator method to modify the raised exception class to
+    `exception_class` but keeps the message and trace intact.
+
+    :param exception_class: class to wrap raised exception with
+    """
     def generic_exception(method):
         def wrapper(*args, **kwargs):
             try:
--- a/bravado_core/formatter.py
+++ b/bravado_core/formatter.py
@@ -78,13 +78,12 @@ class SwaggerFormat(namedtuple('SwaggerFormat',
 
     :param format: Name for the user format.
     :param to_python: function to unmarshal a value of this format.
-                      Eg. lambda str: base64.b64decode(str)
+                      Eg. lambda val_str: base64.b64decode(val_str)
     :param to_wire: function to marshal a value of this format
-                    Eg. lambda val: base64.b64encode(val)
-    :param validate: function to validate the marshalled value. The method
-                     should raise Exception if the value doesn't conform to the
-                     format. `bravado-core` re-raises the exception with
-                     :class:`bravado_core.exception.SwaggerValidationError`.
+                    Eg. lambda val_py: base64.b64encode(val_py)
+    :param validate: function to validate the marshalled value. Raises
+                     :class:`bravado_core.exception.SwaggerValidationError`
+                     if value does not conform to the format.
     :param description: Short description of the format and conversion logic.
     """
 
--- a/bravado_core/formatter.py
+++ b/bravado_core/formatter.py
@@ -3,7 +3,8 @@ Support for the 'format' key in the swagger spec as outlined in
 https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#dataTypeFormat
 """
 
-import collections
+import warnings
+from collections import namedtuple
 
 import six
 import dateutil.parser
@@ -12,6 +13,8 @@ from bravado_core import schema
 
 if six.PY3:
     long = int
+
+NO_OP = lambda x: None
 
 
 def to_wire(spec, value):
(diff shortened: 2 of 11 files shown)


bravado_core/exception.py
```diff
@@ -22,3 +27,17 @@ class SwaggerSchemaError(SwaggerError):
     """Raised when an error is encountered during processing of a SwaggerSchema.
     """
     pass
+
+
+def wrap_exception(exception_class):
```
analogue: ``wrap`` implies that the original exception is intact and available/reachable. Is this the case or am I missing something?

bravado_core/formatter.py
```diff
@@ -32,9 +24,8 @@ def to_wire(spec, value):
     """
     if value is None or not schema.has_format(spec):
         return value
-
-    to_wire, _, _ = _formatters[schema.get_format(spec)]
-    return to_wire(value)
+    formatter = get_format(schema.get_format(spec))
```
analogue: I'm a little concerned that this will now silently no-op for a ```format``` that hasn't been registered. This is different than the previous behavior.  What are the consequences of this change?

bravado_core/formatter.py
```diff
@@ -47,46 +38,77 @@ def to_python(spec, value):
...
+    Register a swagger_format to a global _formatters cache.
 
-if six.PY3:
-    long = int
+    :type format: string
+    :param to_wire: single argument callable that converts a value to wire
+        format
+    :param to_python: single argument callable that converts a value to python
+    :param description: useful description
+    """
```
analogue: Update docstrings

bravado_core/formatter.py
```diff
@@ -47,46 +38,77 @@ def to_python(spec, value):
...
         return value
+    formatter = get_format(schema.get_format(spec))
+    return formatter.to_python(value) if formatter else value
 
-    _, to_python, _ = _formatters[schema.get_format(spec)]
-    return to_python(value)
 
+def register_format(swagger_format):
+    """
+    Register a swagger_format to a global _formatters cache.
```
analogue: This being part of the public interfaces, do callers really need to know about ```global _formatters```? How is this a cache?

bravado_core/formatter.py
```diff
@@ -47,46 +38,77 @@ def to_python(spec, value):
...
+    _formatters[swagger_format.format] = swagger_format
+
+
+def get_format(format):
+    return _formatters.get(format)
+
+
+# validate should check the correctness of `wire` value
+SwaggerFormat = collections.namedtuple(
+    'SwaggerFormat', 'format to_python to_wire validate description')
```
analogue: document attrs and expectations

bravado_core/exception.py
prat0318: `wrap` in the sense it keeps the message and the trace of the initial exception intact and just the exception Class is changed.

bravado_core/formatter.py
```diff
@@ -47,46 +38,77 @@ def to_python(spec, value):
...
+        to_wire=lambda b: b if isinstance(b, str) else str(b),
+        to_python=(
+            lambda s: s if isinstance(s, str) else str(s)),
+        validate=lambda x: None,  # jsonschema validates string
+        description='Converts [wire]string:byte <=> python byte'),
+    'date': SwaggerFormat(
+        format='date',
+        to_wire=lambda d: d.isoformat(),
+        to_python=lambda d: dateutil.parser.parse(d).date(),
+        validate=lambda x: None,  # jsonschema validates date
```
analogue: Maybe a little clearer since it is used a bunch of times.
```python
NO_OP = lambda x: None
...
validate=NO_OP
```

bravado_core/validate.py
```diff
@@ -28,12 +31,20 @@ def validate_schema_object(spec, value):
             obj_type, value))
 
 
+@wrap_exception(ValidationError)
```
analogue: Seems unusual to transform all exceptions. I would imagine the implementation of `validate` would raise a specific exception for validation failure.  

Transforming to a  `jsonschema.ValidationError` seems odd since this really has nothing to do with jsonschema validation. Alternatives?

bravado_core/formatter.py
prat0318: No, callers need not know about _formatters. Will modify the docstring.

bravado_core/formatter.py
prat0318: ack.

bravado_core/formatter.py
prat0318: ack

tests/validate/validate_user_format_test.py
```diff
@@ -0,0 +1,42 @@
...
+        validate=base64.b64decode,
+        description='Base64')
+    register_format(base64_format)
+
+
+def test_success_validate(base64_spec, register_base64_format):
+    try:
+        validate_user_format(base64_spec, to_wire(base64_spec, b'darwin'))
+    finally:
+        del bravado_core.formatter._formatters['base64']
```
analogue: You can use https://pytest.org/latest/fixture.html#fixture-finalization-executing-teardown-code instead

tests/validate/validate_user_format_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_failure_validate_gets_wrapped(base64_spec, register_base64_format):
+    try:
+        validate_user_format(base64_spec, b'darwin')
+    except ValidationError as e:
+        assert 'Incorrect padding' == e.message
+    finally:
+        del bravado_core.formatter._formatters['base64']
```
analogue: ditto

bravado_core/validate.py
prat0318: This was to make a client handle only a single error for swagger validation i.e. `jsonschema.ValidationError`. Other alternative is to wrap `Swagger20Validator(spec).validate(value)` and `validate_user_format` both with `bravado-core.exception.SwaggerError` or `...SwaggerMappingError`

tests/validate/validate_user_format_test.py
prat0318: ack.

bravado_core/formatter.py
prat0318: The consequences are if user forgets to register a user defined format, then the server can still serve the request treating the marshal/unmarshal conversions as a no-op. We can add a `log.warn` or `log.error` saying that the format is not registered.

analogue: I like the warnings.  How about using the built in python support for this? Otherwise, there will be way too much log output. https://docs.python.org/2/library/warnings.html

bravado_core/validate.py
analogue: Yeah, not ideal.  Limiting the validation contract to two exceptions may be the best compromise at this point.  The existing `jsonschema.ValidationError` and a new `SwaggerValidationError` which extends `SwaggerMappingError` since it is a specialization of  `"Raised when an error is encountered during processing of a request or a response."`

bravado_core/formatter.py
prat0318: Agreed, but the default behavior of warnings will be to send output to `stderr` which i think will just get ignored on stage/prod?

analogue: This sort of issue is hard to surface in a way that developers will pay attention to. I think a developer running a service in dev is sufficient (well, internally at least).  

I'm more concerned about the warning surfacing once which is preferred vs on every single invocation.

prat0318: ack.

prat0318:
Added CR suggestions.

prat0318:
Some more review changes ^.

analogue:
lgtm  \o/

PR #36, opened 2015-07-26
Add a convenience method for easily iterating over all operations of a spec

sjaensch:
We wrote something similar to this since we needed a list of all operations (paths with HTTP methods). We used this both to create a mock server (a server that returns predefined responses when receiving HTTP request) as well as to test all endpoints. The cast of HTTP method to uppercase is for convenience as well since almost all server-side software expects the method name to be in uppercase.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+4.2.5 (2016-07-27)
+------------------
+- Add basepython python2.7 for flake8, docs, and coverage tox commands
+
 4.2.4 (2016-07-26)
 ------------------
 - coverage v4.2 was incompatible and was breaking the build. Added --append for the fix.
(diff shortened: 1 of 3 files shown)


bravado_core/spec.py
```diff
@@ -123,6 +123,21 @@ def get_op_for_request(self, http_method, path_pattern):
...
+        for server-side implementations.
+
+        :returns: an iterator that returns two-tuples containing the HTTP method
+            and path_name. HTTP method is in uppercase.
+        """
+        values_attr = 'itervalues' if hasattr(dict, 'itervalues') else 'values'
+        for resource in getattr(self.resources, values_attr)():
+            for op in getattr(resource.operations, values_attr)():
+                yield (op.http_method.upper(), op.path_name)
+
```
analogue: I'm not sure if this is something that would be useful outside your specific usage. Is there currently more than one user of this library that is duplicating the above code?

PR #38, opened 2015-08-06
Fix bravado_core module docstrings

prat0318:


docs/source/bravado_core.rst
```diff
@@ -0,0 +1,142 @@
...
+Submodules
+----------
+
+bravado_core.content_type module
+--------------------------------
+
+.. automodule:: bravado_core.content_type
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
dnephin: I usually only use `:members:` here.  I would consider undocumented things to not be public, so shouldn't show up in these docs, and I don't think we have any inheritance that is significant to the docs.

analogue:
lgtm

docs/source/bravado_core.rst
prat0318: I think sometimes members go undocumented by mere ignorance, and not because of intentionally they are private. Also question, Is there a way to remove any of these three when running `sphinx-apidoc`

dnephin: Not that I know of. I think sphinx-apidoc is only intended to generate a starting point for you.  From there you can edit and add as you want.

A lot of the time each of these sections will go into a separate file.

PR #39, opened 2015-08-07
Use a more specific exception when a response can't be mapped to the schema

analogue:
Needed for https://github.com/Yelp/bravado/issues/153

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -3,7 +3,7 @@ from six import iteritems
 from bravado_core.content_type import APP_JSON
 from bravado_core.unmarshal import unmarshal_schema_object
 from bravado_core.validate import validate_schema_object
-from bravado_core.exception import SwaggerMappingError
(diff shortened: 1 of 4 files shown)


bravado_core/response.py
```diff
@@ -3,7 +3,7 @@
 from bravado_core.content_type import APP_JSON
 from bravado_core.unmarshal import unmarshal_schema_object
 from bravado_core.validate import validate_schema_object
-from bravado_core.exception import SwaggerMappingError
+from bravado_core.exception import SwaggerMappingError, MatchingResponseNotFound
```
prat0318: Nit: MatchingResponseNotFound to come earlier than SwaggerMappingError

prat0318:
shipit

bravado_core/response.py
analogue: ack

PR #41, opened 2015-08-11
Dereference elements from a list. Fixes #40.

### LINKED ISSUE #40: jsonref dereference happening two times, leading to failures.
Following `swagger.json` fails:

``` json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "Title was not specified"
  },
  "tags": [
    {
      "name": "hours",
      "description": "Retrieve business hours."
    }
  ],
  "paths": {
    "/hours/{biz_id}": {
      "get": {
        "produces": ["application/json"],
        "responses": {
          "200": {
            "$ref": "biz_id_param.json"
          },
          "405": {
            "description": "Invalid input"
          }
        },
        "description": "",
        "tags": [
          "hours"
        ],
        "summary": "Get happy hour times by biz_id",
        "operationId": "getHours",
        "parameters": [
          {
              "$ref": "file:///home/prat0318/param.json"
          }
        ]
      }
    }
  },
  "host": "localhost:1200",
  "basePath": "/",
  "schemes": [
    "http"
  ],
  "definitions": { }
}
```

and `param.json` being:
``` json
{
            "in": "path",
            "description": "Business ID",
            "required": true,
              "name": "biz_id",
            "type": "string"
}
```
(To run the example, MENTION full `file://` address of `param.json` in above `swagger.json`)

The issue being dict gets dereferenced twice, once during `from_dict` [init](https://github.com/Yelp/bravado-core/blob/49a322f5aec84f7c50409c136f82888ed96b26f0/bravado_core/spec.py#L88) and later when it calls build, and then [validate_spec](https://github.com/Yelp/swagger_spec_validator/blob/master/swagger_spec_validator/validator20.py#L41).

Note: The error does not occur in all cases, but in some specific cases.

prat0318:
- Fixes #40
- Dereference from dict was happening but not from list.
- If `assert not isinstance(json_obj['foo'][0], jsonref.JsonRef)` was added at L93 to `tests/spec/replace_jsonref_proxies_test.py`, it would fail. It passes now.

### DIFF
--- a/tests/spec/replace_jsonref_proxies_test.py
+++ b/tests/spec/replace_jsonref_proxies_test.py
@@ -20,10 +20,10 @@ def test_dict():
         }
     }
     json_obj = jsonref.loads(json.dumps(d))
-    assert isinstance(json_obj['foo'], jsonref.JsonRef)
+    assert type(json_obj['foo']) == jsonref.JsonRef
 
     replace_jsonref_proxies(json_obj)
-    assert isinstance(json_obj['foo'], dict)
+    assert type(json_obj['foo']) == dict
 
     assert d['definitions']['user'] == json_obj['foo']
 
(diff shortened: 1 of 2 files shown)


tests/spec/replace_jsonref_proxies_test.py
```diff
@@ -23,6 +23,7 @@ def test_dict():
     assert isinstance(json_obj['foo'], jsonref.JsonRef)
 
     replace_jsonref_proxies(json_obj)
+    assert not isinstance(json_obj['foo'], jsonref.JsonRef)
     assert isinstance(json_obj['foo'], dict)
```
analogue: Why not tighten it up to match the actual instance type since ```isinstance``` is true for subclassses as well. I  assume thats is why the original assertion didn't catch this? Something like 

```assert type(json_obj['foo']) == <actual type>```

prat0318: Ack. Changed all the `isinstance` checks to `type` checks.

analogue:
:+1:

Issue #48, opened 2015-09-24
Validation of user defined formats only works for primitive schemas


analogue:
Given:
```python
def validate_palindrome(x):
    if x != ''.join(reversed(x)):
        raise SwaggerValidationException('%s is not a palindrome' % x)

palindrome_format = SwaggerFormat(
    format='palindrome',
    to_wire=lambda x: x,
    to_pyton=lambda x: x,
    validate=validate_palindrome)
```
This works for schema:
```json
{
    "name": "foo",
    "type": "string",
    "format": "palindrome"
}
```
But doesn't work for nested primitives:
```json
{
    "type": "object",
    "properties": {
        "foo": {
            "type": "string", 
            "format": "palindrome"
        }
    }
}
```
This is because validation is outsourced wholesale to `Swagger20Validator` which is not aware of the registered `SwaggerFormat`s.  Few options here:
1. Figure out how to hook the SwaggerFormat validation functions into Swagger20Validator which in turn uses `jsonschema` (best)
2. Make another pass over the schema post-Swagger20Validator and find/validate every `format` found (ick, a hack!)
3. Post-pone the validation for user defined formats until unmarshalling and do it then (double ick, a nastier hack!)

PR #49, opened 2015-09-28
Fix validation for user-defined formats

analogue:
See #48 

This farms out user-defined format validation to `jsonschema`.


bravado_core/formatter.py
```diff
@@ -88,6 +110,50 @@ class SwaggerFormat(namedtuple('SwaggerFormat',
...
+    Build and cache a :class:`jsonschema.FormatChecker` for validating
+    user-defined Swagger formats.
+
+    :rtype: :class:`jsonschema.FormatChecker`
+    """
+    global _format_checker
+    if _format_checker is None:
+        _format_checker = FormatChecker()
+        for swagger_format in _user_defined_formats:
+            _format_checker.checkers[swagger_format.format] = (
```
prat0318: Instead use [the API](http://python-jsonschema.readthedocs.org/en/latest/validate/#jsonschema.FormatChecker.checkers) available for adding a checker?

bravado_core/formatter.py
```diff
@@ -88,6 +110,50 @@ class SwaggerFormat(namedtuple('SwaggerFormat',
...
+
+def get_format_checker():
+    """
+    Build and cache a :class:`jsonschema.FormatChecker` for validating
+    user-defined Swagger formats.
+
+    :rtype: :class:`jsonschema.FormatChecker`
+    """
+    global _format_checker
+    if _format_checker is None:
```
prat0318: Won't global format_checker increase the pain described in #50?

bravado_core/formatter.py
```diff
@@ -88,6 +110,50 @@ class SwaggerFormat(namedtuple('SwaggerFormat',
     """
 
 
+def return_true_wrapper(validate_func):
+    """Decorator for the SwaggerFormat.validate function to always return True.
+
+    The contract for `SwaggerFormat.validate` is to raise an exception
+    when validation fails. However, the contract for jsonschema's
+    validate function is to raise an exception or return True. This wrapper
+    bolts-on the `return True` part.
```
prat0318: Also link it to the jsonschema doc : http://python-jsonschema.readthedocs.org/en/latest/validate/#jsonschema.FormatChecker

bravado_core/formatter.py
```diff
@@ -46,18 +48,38 @@ def to_python(spec, value):
...
     """
     global _formatters
     _formatters[swagger_format.format] = swagger_format
 
+    # Need to maintain a separate list of UDFs for jsonschema validation
+    global _user_defined_formats
+    _user_defined_formats.append(swagger_format)
+
+
+def unregister_format(swagger_format):
```
prat0318: When/why will an application call this function?

docs/source/changelog.rst
```diff
@@ -0,0 +1 @@
+../../CHANGELOG.rst
```
prat0318: does this work?

tests/validate/validate_array_test.py
```diff
@@ -54,3 +55,45 @@ def test_uniqueItems_false(int_array_spec):
...
+        'type': 'array',
+        'items': {
+            'type': 'string',
+            'format': 'ipv4',
+        }
+    }
+    request_body = ['not_an_ip_address']
+    with pytest.raises(ValidationError) as excinfo:
+        validate_array(ipaddress_array_spec, request_body)
+    assert "'not_an_ip_address' is not a 'ipv4'" in str(excinfo.value)
```
prat0318: but where is ipv4 format registered?

bravado_core/formatter.py
analogue: The API only provides a function decorator.

bravado_core/formatter.py
analogue: Yes. Better to fix this bug in a backwards compatible way before releasing a non-backward compatible version.

bravado_core/formatter.py
analogue: Don't really think this is necessary. FormatChecker is integral to the validation of user-defined formats. Any developer looking for more info is going to easily be able to navigate to the source. Furthermore, this isn't a part of the public API.

bravado_core/formatter.py
analogue: See unit tests

docs/source/changelog.rst
analogue: Yes

tests/validate/validate_array_test.py
analogue: As the test method name implies, it is a builtin format.

bravado_core/formatter.py
prat0318:
```python
validator = return_true_wrapper(swagger_format.validate)
_format_checker.checks(
    swagger_format.format, raises=(SwaggerValidationError,))(validator)
```
This way there is no need to modify `checkers` attribute directly.

bravado_core/formatter.py
prat0318: I am thinking if there is an alternate way than introducing global _format_checker and _user_defined_formats.

bravado_core/formatter.py
prat0318: fine with me.

bravado_core/formatter.py
prat0318: My question was from the point of view of actual use-case. Pyramid registers user-formats during init of pyramid_swagger, but it need not unregister them at any time. Is it meaningful from client side use-case?

docs/source/changelog.rst
prat0318: cool.

bravado_core/formatter.py
analogue: What would that way be?

bravado_core/formatter.py
analogue: I'd say no. Moving this function to the unit test might be an improvement, but then the scope of the function is unnecessarily limited. What is the critical issue here?

bravado_core/formatter.py
analogue: Ah, the example makes it clear.

bravado_core/formatter.py
prat0318: We can then get rid of _user_defined_formats. Addition to _format_checker can happen directly with register_format (and now we do not care about rebuilding it).

bravado_core/formatter.py
prat0318: i can only think of a way to get rid of _udf (described above). i guess we have to live with _format_checker till _formatters and _format_checker are associated with Spec (as described in #50)

tests/validate/validate_array_test.py
prat0318: oh there are some jsonschema built-ins. they look useful and the user need not register the format with definitions for these to work.

Issue #50, opened 2015-09-29
User-defined formats should be tied to a Spec (not global)


analogue:
Currently, user-defined formats are defined globally.  This is not ideal when a python client process interfaces with more than one Swagger service with a different set of user-defined formats applying to each service.  In the worst case, the same format name may be used with differing implementations.

Associate user-defined formats with a `bravado_core.spec.Spec` instead.

analogue:
Released in 3.0.0

PR #51, opened 2015-10-02
Register user-defined formats with a Spec instead of being global

analogue:
See #50 

User-defined formats are now registered like so:
```python
swagger_spec = Spec.from_dict(...)
swagger_spec.register_format(my_super_duper_format)
```

Additional changes:
- Deprecated `bravado_core.request.RequestLike` and renamed to `bravado_core.request.IncomingRequest`.
- Added `make docs` target
- Generated sphinx docs are no longer a verbatim copy of bravado's docs.

Coverage remains at 97%


bravado_core/formatter.py
```diff
@@ -19,93 +17,51 @@
 NO_OP = lambda x: None
 
 
-def to_wire(spec, value):
+def to_wire(swagger_spec, primitive_spec, value):
     """Converts a python primitive or object to a reasonable wire
```
prat0318: Should it be just ".. python primitive to a ..."

prat0318:
:+1:

bravado_core/formatter.py
analogue: It can be an object. For example, an `ObjectID` instance that gets marshalled to a string.

Issue #45, opened 2015-08-13
Also tag models with external references.


prat0318:
Currently only models present in `#/definitions` get tagged with `x-model: <model name>` and get used during marshalling/unmarshalling for type conversion to a python class versus a dict. Other models referred using `file://` or `http://` should also be treated in the same manner.

There can be cases where `$ref` refers to a complete file, like shown [here](https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md#definitions):
```json
{
  "description": "A simple model",
  "type": "object",
  "properties": {
    "id": {
      "type": "integer",
      "format": "int64"
    },
    "tag": {
      "description": "a complex, shared property.  Note the absolute reference",
      "$ref": "https://my.company.com/definitions/Tag.json"
    }
  }
}
```
These kind of `ref`s can be ignored, as they do not contain a `Tag` name, to associate them with.

analogue:
Fixed in 3.0.1

PR #52, opened 2015-10-06
Tag models in external refs with 'x-model' automagically

analogue:
See #45

All models for the usage scenarios we've seen (local and external refs) are now tagged so schema authors don't have to.


bravado_core/spec.py
```diff
@@ -1,19 +1,21 @@
...
 from six.moves.urllib import parse as urlparse
 from swagger_spec_validator import validator20
+
 from bravado_core import formatter
 from bravado_core.exception import SwaggerSchemaError, SwaggerValidationError
 from bravado_core.formatter import return_true_wrapper
-
-from bravado_core.model import build_models
-from bravado_core.model import tag_models
+from bravado_core.model import annotate_with_xmodel_callback, \
```
prat0318: nit: instead use 
```
import (...,
                                ...)
```

bravado_core/model.py
```diff
@@ -14,22 +16,87 @@
 # differentiated from 'object' types.
 MODEL_MARKER = 'x-model'
 
+RE_MODEL_NAME = re.compile(r"""
+    [\w.]*              # Skip filename if specified
+    \#/definitions/     # match `#/definitions/`
```
prat0318: I am not a huge fan of hard-coding the `#/definitions/` part. As in the [2nd example of this section](https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md#definitions) they ref with: `"$ref": "https://my.company.com/definitions/models.json#/models/Tag"` which won't follow the regex. I have not thought about a correct way yet (everything after the last `/` in the string, may be).

prat0318:
not gone through completely. will go through the callbacks defn. tomo.

bravado_core/spec.py
analogue: Fixed

bravado_core/model.py
analogue: See the recently added [docs](https://github.com/Yelp/bravado-core/commit/fcf0d3c6f6fe4fbb7499daf5c3e276ebac2b9036#diff-fa72ede42c4474d2a46216e93715104cR1) which outline the supported discovery mechanisms and cover our immediate needs.

prat0318: On second thoughts, this kind of best effort search strategy can lead to subtle bugs as discovered by @nirmaldm. Suppose service developer X was working on a single `swagger.json` file. And all the clients started using models in their logic. Now when it grew big, unknowingly, X refactored it such that definitions do not exactly follow the `#/definitions/format`. There were no errors from service side, so X deployed it. But then, all clients will break, seeing a dict response instead of a model. This becomes worse when the client users have explicitly mentioned `use_models` to be `True`.

analogue: A bit far fetched and a bit of an extreme edge case: "X refactored it such that definitions do not exactly follow the #/definitions/format". That is what documentation is for :smile: 

I'm trying to understand what the default has to do with it. So clients have code that uses the model, the model disappears because it can't be discovered anymore (even though the documentation spells out explicitly how models are found). What does the default have to do with this failure?

bravado_core/model.py
```diff
@@ -14,22 +16,87 @@
...
+    :param key: the key of the object in the container to inspect
+    :type key: string if container is a dict, int if container is a list
+    """
+    jsonref_proxy = container[key]
+    if not isinstance(jsonref_proxy, jsonref.JsonRef):
+        return
+
+    model = jsonref_proxy.__subject__
+    if is_model(model):
+        model_name = model['x-model']
```
prat0318: s/'x-model'/MODEL_MARKER

bravado_core/model.py
```diff
@@ -14,22 +16,87 @@
...
     """
-    models = {}
-    for model_name, model_spec in iteritems(definitions_spec):
-        # make models available under both simple name and $ref style name
-        # - Pet <-- TODO: remove eventually
-        # - #/definitions/Pet
-        models[model_name] = create_model_type(model_name, model_spec)
-        models['#/definitions/{0}'.format(model_name)] = models[model_name]
-    return models
+    if key != MODEL_MARKER:
```
prat0318: i am not sure i follow this. `key` is supposed to be either index (0,1,2..) or dict key like ('Pet', 'Tag'..). So, when will the key take the value `x-model`? Is it auto-generated somewhere before?

prat0318:
lgtm. I would admit though complexity of the logic has grown manyfolds after bringing `jsonref` into the scenes.

bravado_core/model.py
analogue: Fixed this and many others.

PR #56, opened 2015-10-12
Don't create an operation out of path parameters

nattofriends:
`parameters` isn't a HTTP method, trying to make an `Operation` out of it will not turn out well.


analogue:
A [unit test](https://github.com/Yelp/bravado-core/blob/master/tests/resource/build_resources_test.py) would be nice!

analogue:
Added unit test in #58

nattofriends:
Sorry, I totally should have had a test to go along with this in the first place. :X

PR #58, opened 2015-10-19
Don't create an operation out of path parameters (with unit test)

analogue:
Builds on PR #56 by adding a unit test.

### DIFF
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -49,7 +49,6 @@ def build_resources(swagger_spec):
     paths = swagger_spec.spec_dict['paths']
     for path_name, path_spec in iteritems(paths):
         for http_method, operation_spec in iteritems(path_spec):
-            
             # parameters that are shared across all operations for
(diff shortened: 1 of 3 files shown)


bravado_core/resource.py
```diff
@@ -49,6 +49,9 @@ def build_resources(swagger_spec):
     paths = swagger_spec.spec_dict['paths']
     for path_name, path_spec in iteritems(paths):
         for http_method, operation_spec in iteritems(path_spec):
+            if http_method == 'parameters':
```
prat0318: It took me a while to understand how `parameters` can get assigned as `http_method`. A comment about this - handling `parameters` key in swagger spec which can be at same level as http methods - can be helpful later.

analogue: Good point. I'll add something.

prat0318:
flake8 complaining :(

prat0318:
fix n ship

Issue #35, opened 2015-07-22
Recursive Model Definition is not correctly handled


laucia:
A simple recursive model definition like this
```
{
    "swagger": "2.0",
    "info": {
        "version": "1.0.0",
        "title": ""
    },
    "paths": {
        "/recursion": {
            "get": {
                "responses": {
                    "200": {
                        "description": "Recurse",
                        "schema": {
                            "$ref": "#/definitions/Recursion"
                        }
                    }
                },
                "parameters": []
            }
        }
    },
    "host": "localhost",
    "basePath": "/",
    "definitions": {
        "Recursion": {
            "properties": {
                "title": {
                    "description": "user facing localized display string",
                    "type": "string"
                },
                "parent": {
                    "description": "parent recursion",
                    "$ref": "#/definitions/Recursion"
                }
            },
            "required": [
                "title"
            ]
        }
    }
}

```

Leads to this exception

```
  File "/usr/local/lib/python2.7/site-packages/bravado_core/spec.py", line 203, in descend
    if is_dict_like(fragment):
  File "/usr/local/lib/python2.7/site-packages/bravado_core/schema.py", line 49, in is_dict_like
    if type(spec) == dict:
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

As nothing in the swagger spec advise against recursive model definitions and the canonical swagger-codegen [handles it](https://github.com/swagger-api/swagger-codegen/issues/728), this should also work in bravado.

analogue:
This is not going to be a simple fix. Just some notes:
- swagger_spec_validator and bravado_core both need this functionality
- jsonschema doesn't support validation of json documents with external refs (local refs are ok) which is a necessary feature
- usage of jsonref will have to be removed entirely since it doesn't support cyclic refs
- all call sites that may encounter a ref need to resolve the ref transparently at runtime
- do some research on how this is handled best in existing libs outside the python ecosystem

analogue:
Fix in 4.0.0

PR #59, opened 2015-10-30
Support for recursive refs

analogue:
See #35 for context.

This PR adds support for recursive $refs. This is not something that can be broken up and added incrementally, hence the large diff. Basically, any read/write operations to ``swagger_spec.spec_dict`` will attempt to dereference a ``$ref`` if present. The ``RefResolver`` from the jsonschema library is used extensively to perform the dereferencing.

Highlights:
- jsonref no longer a dependency
- petstore spec in test-data updated to latest and greatest
- ~~swagger spec validation (not request/response validation) is temporarily disabled until swagger-spec-validator is also updated to support recursive refs~~
- ``bravado_core.spec.Spec.deref(..)`` and ``bravado_core.spec.post_process_spec(...)`` are temporary. They will use similiar functions that will reside in swagger-spec-validator.
- Minor doc updates along the way
- ``bravado_core.operation.build_params(..)`` fixed to support overriding same parameter at the path level
- Un-qualified refs such as ``{"$ref": "Category"}`` no longer supported  (follwing the lead of the petstore example)
- Automatic model tagging is only supported in the top-level spec file (determining fully qualified json pointer path is a little complicated with traversing non-root level refs)
- unit test suite passes


bravado_core/marshal.py
```diff
@@ -22,11 +19,14 @@ def marshal_schema_object(swagger_spec, schema_object_spec, value):
...
     :type value: int, long, string, unicode, boolean, list, dict, Model type
+
     :return: marshaled value
     :rtype: int, long, string, unicode, boolean, list, dict
     :raises: SwaggerMappingError
     """
-    obj_type = schema_object_spec['type']
+    deref = swagger_spec.deref
+    schema_object_spec = deref(schema_object_spec)
+    obj_type = schema_object_spec.get('type')
```
prat0318: isn't 'type' a necessary key?

bravado_core/marshal.py
```diff
@@ -3,17 +3,14 @@
...
 from bravado_core.model import is_model, MODEL_MARKER
-from bravado_core.schema import (
-    is_dict_like,
-    is_list_like,
-    SWAGGER_PRIMITIVES,
-    get_spec_for_prop
-)
+from bravado_core.schema import is_dict_like
+from bravado_core.schema import is_list_like
+from bravado_core.schema import SWAGGER_PRIMITIVES
```
prat0318: nit: ordering

bravado_core/model.py
```diff
@@ -1,105 +1,73 @@
...
+    if len(path) < 2 or path[-2] != 'definitions':
         return
+    deref = swagger_spec.deref
+    model_name = key
+    model_spec = deref(container.get(key))
 
-    ref_target = jsonref_proxy.__reference__['$ref']
-    match = RE_MODEL_NAME.match(ref_target)
-    if match is None:
+    if deref(model_spec.get('type')) != 'object':
```
prat0318: Isn't mode_spec already all `deref`ed? do we need another `deref()`?

bravado_core/resource.py
```diff
@@ -45,46 +45,48 @@ def build_resources(swagger_spec):
...
+        resources[tag] = Resource(tag, ops)
     return resources
 
 
 class Resource(object):
+    """A Swagger resource is associated with multiple operations.
+
+    :param name: resource name
+    :type name: str
+    :param operations: operations associated with this resource (by tag)
```
prat0318: s/operations/ops

prat0318:
there has to be all print cleanups. looks good otherwise. i haven't looked much in the tests though.

analogue:
Blocked on https://github.com/Yelp/swagger_spec_validator/pull/41 getting merged and released

bravado_core/model.py
analogue: ``deref`` only applies to the value passed in.

PR #61, opened 2015-11-26
Add support for marshalling dict-like objects.

obmarg:
The `is_dict_like` function in `schema.py` was just checking if the
object's type was a dict.  This causes it to treat any dict subclass as
a model, which fails because dict subclasses are not models.

Updating to check if the item is an instance of `collections.Mapping`
(or `collections.abc.Mapping` on python 3) fixes this.

### DIFF
--- a/bravado_core/schema.py
+++ b/bravado_core/schema.py
@@ -1,5 +1,10 @@
 from bravado_core.exception import SwaggerMappingError
 
+try:
+    from collections import Mapping
+except ImportError:
+    from collections.abc import Mapping
+
 # 'object' and 'array' are omitted since this should really be read as
 # "Swagger types that map to python primitives"
 SWAGGER_PRIMITIVES = (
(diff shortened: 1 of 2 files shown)


bravado_core/schema.py
```diff
@@ -1,5 +1,10 @@
 from bravado_core.exception import SwaggerMappingError
 
+try:
+    from collections import Mapping
+except ImportError:
+    from collections.abc import Mapping
```
asottile: fwiw, I doubt this ImportError will ever be hit (even in python3):

```python
# head -n 8 /usr/lib/python3.5/collections/__init__.py
__all__ = ['deque', 'defaultdict', 'namedtuple', 'UserDict', 'UserList',
            'UserString', 'Counter', 'OrderedDict', 'ChainMap']

# For backwards compatibility, continue to make the collections ABCs
# available through the collections module.
from _collections_abc import *
import _collections_abc
__all__ += _collections_abc.__all__
```

analogue:
Will fix the extraneous import logic post-merge.

analogue:
@obmarg Thanks! :+1:

Issue #47, opened 2015-09-24
Add allow_null option


jpizarrom:
Add allow_null option like
https://github.com/Yelp/bravado-core/blob/632a0f9a9ec0a3fe5c996e932b01fec9a1c933d8/docs/source/quickstart.rst#advanced-options

jpizarrom:
It's posible to use x-nullable like propose in https://github.com/swagger-api/swagger-spec/issues/229

The use of x-nullable respect the spec https://github.com/swagger-api/swagger-spec/blob/d79c205485d702302003d4de2f2c980d1caf10f9/versions/2.0.md#patterned-objects

example of a property
```
"next": {
  "type": "string",
  "x-nullable": true
 }
```

jpizarrom:
https://github.com/admetricks/bravado-core/commit/9360ff2e7fcee14069b2f4b0ae267e3047901586

andreashug:
This would be great. I'm using Swagger to build a client for a third-party API so changing it to match with the specification is not an option.

Would you accept a pull request with tests and documentation?

analogue:
I'm a little concerned that the next version of the Swagger Spec will start to support this behavior but in a possibly different way.  Anyone know what the plans are for this feature in the working group?

Maybe @webron can chime in here?

webron:
I'm afraid I don't have a clear answer to that as the official discussions haven't started. Right now, it's just the community chiming in. Haven't thoroughly investigated it, but it is possible we'll have a property that's `isNullable` or something like that which would default to either true or false.

analogue:
@andreashug Feel free to submit a PR

sjaensch:
This is fixed with the support for the property `x-nullable`.

PR #64, opened 2016-01-04
Allow null option for properties

andreashug:
Proposal for solving #47

The Swagger specification in version 2 doesn't support 'null' as
property value. This adds a vendor extension 'x-nullable' which
allows properties to be 'null' if set to 'true'.

### DIFF
--- a/tests/swagger20_validator/type_validator_test.py
+++ b/tests/swagger20_validator/type_validator_test.py
@@ -34,3 +34,25 @@ def test_validate_when_not_a_parameter_schema(m_draft4_type_validator,
             string_schema)
     list(type_validator(minimal_swagger_spec, *args))
     m_draft4_type_validator.assert_called_once_with(*args)
+
+
+@patch('jsonschema._validators.type_draft4')
+def test_skip_when_nullable_property_schema_and_value_is_None(
+        m_draft4_type_validator, minimal_swagger_spec):
+    param_schema = {'name': 'foo', 'x-nullable': True, 'type': 'string'}
+    type_validator(
+        minimal_swagger_spec,
+        validator=None,
+        types=param_schema['type'],
+        instance=None,  # property value
+        schema=param_schema)
+    assert m_draft4_type_validator.call_count == 0
+
+
+@patch('jsonschema._validators.type_draft4')
+def test_validate_when_not_nullable_property_schema_and_value_is_None(
+        m_draft4_type_validator, minimal_swagger_spec):
+    param_schema = {'name': 'foo', 'x-nullable': False, 'type': 'string'}
+    args = (None, param_schema['type'], None, param_schema)
+    type_validator(minimal_swagger_spec, *args)
+    m_draft4_type_validator.assert_called_once_with(*args)
--- a/tests/swagger20_validator/type_validator_test.py
+++ b/tests/swagger20_validator/type_validator_test.py
@@ -39,20 +39,20 @@ def test_validate_when_not_a_parameter_schema(m_draft4_type_validator,
 @patch('jsonschema._validators.type_draft4')
 def test_skip_when_nullable_property_schema_and_value_is_None(
         m_draft4_type_validator, minimal_swagger_spec):
-    param_schema = {'name': 'foo', 'x-nullable': True, 'type': 'string'}
-    type_validator(
+    prop_schema = {'x-nullable': True, 'type': 'string'}
+    list(type_validator(
         minimal_swagger_spec,
         validator=None,
-        types=param_schema['type'],
+        types=prop_schema['type'],
         instance=None,  # property value
-        schema=param_schema)
+        schema=prop_schema))
     assert m_draft4_type_validator.call_count == 0
 
 
 @patch('jsonschema._validators.type_draft4')
 def test_validate_when_not_nullable_property_schema_and_value_is_None(
         m_draft4_type_validator, minimal_swagger_spec):
-    param_schema = {'name': 'foo', 'x-nullable': False, 'type': 'string'}
-    args = (None, param_schema['type'], None, param_schema)
-    type_validator(minimal_swagger_spec, *args)
+    prop_schema = {'x-nullable': False, 'type': 'string'}
+    args = (None, prop_schema['type'], None, prop_schema)
+    list(type_validator(minimal_swagger_spec, *args))
     m_draft4_type_validator.assert_called_once_with(*args)
--- a/tests/unmarshal/unmarshal_nullable_test.py
+++ b/tests/unmarshal/unmarshal_nullable_test.py
@@ -5,60 +5,66 @@ from bravado_core.validate import validate_schema_object
 from jsonschema.exceptions import ValidationError
 
 
-@pytest.mark.parametrize('value', ['x', None])
-@pytest.mark.parametrize('nullable', [True, False])
-def test_unmarshal_with_primitive(empty_swagger_spec, value, nullable):
-    """If the value is a primitive, validation should pass if
-    `x-nullable` is `True` and the value is `None`. `required` is not
-    possible in this scenario.
+"""If the value is a primitive, validation should pass if
+`x-nullable` is `True` and the value is `None`. `required` is not
+possible in this scenario.
 
-    +---------------------+--------------+
-    | x-nullable == False | 'x'  -> pass |
-    |                     | None -> fail |
-    +---------------------+--------------+
-    | x-nullable == True  | 'x'  -> pass |
-    |                     | None -> pass |
-    +---------------------+--------------+
-    """
++---------------------+------------------+
+| x-nullable == False | 'x'  -> pass (1) |
+|                     | None -> fail (2) |
++---------------------+------------------+
+| x-nullable == True  | 'x'  -> pass (3) |
+|                     | None -> pass (4) |
++---------------------+------------------+
+"""
+
+
+@pytest.mark.parametrize(['nullable', 'value'],
+                         [(False, 'x'), (True, 'x'), (True, None)])
+def test_unmarshal_with_primitive_pass(empty_swagger_spec, value, nullable):
+    """Test scenarios in which validation should pass: (1), (3), (4)"""
     content_spec = {
         'type': 'string',
         'x-nullable': nullable,
     }
-
-    try:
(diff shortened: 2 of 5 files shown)


msander:
+1

analogue:
Thanks for the PR. Haven't had time to review yet so it will probably be next week.

tests/swagger20_validator/type_validator_test.py
```diff
@@ -34,3 +34,25 @@ def test_validate_when_not_a_parameter_schema(m_draft4_type_validator,
             string_schema)
     type_validator(minimal_swagger_spec, *args)
     m_draft4_type_validator.assert_called_once_with(*args)
+
+
+@patch('jsonschema._validators.type_draft4')
+def test_skip_when_nullable_property_schema_and_value_is_None(
+        m_draft4_type_validator, minimal_swagger_spec):
+    param_schema = {'name': 'foo', 'x-nullable': True, 'type': 'string'}
```
analogue: I think you mean for this to be `prop_schema` with no `name` key.

tests/swagger20_validator/type_validator_test.py
```diff
@@ -34,3 +34,25 @@ def test_validate_when_not_a_parameter_schema(m_draft4_type_validator,
...
+        types=param_schema['type'],
+        instance=None,  # property value
+        schema=param_schema)
+    assert m_draft4_type_validator.call_count == 0
+
+
+@patch('jsonschema._validators.type_draft4')
+def test_validate_when_not_nullable_property_schema_and_value_is_None(
+        m_draft4_type_validator, minimal_swagger_spec):
+    param_schema = {'name': 'foo', 'x-nullable': False, 'type': 'string'}
```
analogue: Same here

analogue:
What is your interpretation of `x-nullable` when `required==True`?  If I'm reading this correctly,  when a property is nullable and required and the value passed is None, then that makes it OK.  Hence, there would have to be changes to `unmarshal_primitive(..)` and `marshal_primitive(..)`

tests/swagger20_validator/type_validator_test.py
andreashug: You are right. I'll change that.

tests/swagger20_validator/type_validator_test.py
andreashug: Same here ;-)

andreashug:
My understanding of the spec is that `required==True` only means the property must exist and `x-nullable==True` allows `null` as value.
 
 | `required==False`  |  `required==True` |
----- | ----- | ------
`x-nullable==False` | `{ }->ok` `{"x": null}->fail`  |  `{ }->fail` `{"x": null}->fail`
`x-nullable==True` | `{ }->ok` `{"x": null}->ok` | `{ }->fail` `{"x": null}->ok`

I'm not sure if this is possible because `unmarshal_object(..)` re-introduces missing properties with `None`. What do you think?

andreashug:
While searching for a good solution, I discovered that JSON Schema allows a list of types - e.g. `
{
    "schema": {
        "type": ["string", "null"]
    }
`
Actually I think this would be a much better way to solve my problem and it works well with `jsonschema`. Unfortunately it's not supported bravado-core yet. I'll try to come up with a proposal so this pull request can be postponed.

wasbazi:
This would be a really nice feature to have! There is a long thread on the OpenAPI-Specification (formerly The Swagger Specification) repository discussing how they want to solve this and leaning towards the `x-nullable` case https://github.com/OAI/OpenAPI-Specification/issues/229

Unfortunately it seems from the discussion that using arrays for type is not intended with the 2.0 spec (which seems to be indicated as feature frozen) https://github.com/OAI/OpenAPI-Specification/issues/229#issuecomment-164587628

Are there things I can do to help get this PR finished?

cyncyncyn:
As @wasbazi says, this would be a nice feature to have!
I also offer my help in order to get this PR merged.
Thanks!

sjaensch:
The CR has an outstanding issue, see https://github.com/Yelp/bravado-core/pull/64#issuecomment-175177203
The proposed solution of going with a list of types as allowed by JSON Schema will not be possible, the Swagger spec authors already clarified that Swagger 2.0 is not supposed to allow a list of types.

normalex:
Can we merge this? people are eager to see it released.

andreashug:
Finally found some time to continue with this.
1) Rebased on recent master
2) Fixed test according to https://github.com/Yelp/bravado-core/pull/64#issuecomment-170170418
3) Integration tests for unmarshaling

The good news is that unmarshaling doesn't seem to be affected (see matrix in https://github.com/andreashug/bravado-core/blob/c581aaba8924afe340f124375015afc9f5743762/tests/unmarshal/unmarshal_nullable_test.py)

The not so good news however is, I'm not sure how marshaling should behave in combination with `x-nullable`. Currently, `marshal_object()` drops keys whose value is `None`, which leads to a ValidationError (required property) -> https://github.com/Yelp/bravado-core/blob/7d13d2512f85983eca086db1eb0ddab797b22499/bravado_core/marshal.py#L129-L130. One might think, it should keep the key if `x-nullable` is `True`. Or we could say that `x-nullable` is only for properties and not accounted in parameter specs.

Any opinions on that?

sjaensch:
Great stuff @andreashug! I'm still travelling, I'll take a closer look at the patch on Tuesday.
Just to make sure I understand the issue you're seeing: If an endpoint expects a parameter of type object that has a `x-nullable` field then we get a ValidationError?

tests/swagger20_validator/type_validator_test.py
```diff
@@ -34,3 +34,25 @@ def test_validate_when_not_a_parameter_schema(m_draft4_type_validator,
...
+        schema=prop_schema))
+    assert m_draft4_type_validator.call_count == 0
+
+
+@patch('jsonschema._validators.type_draft4')
+def test_validate_when_not_nullable_property_schema_and_value_is_None(
+        m_draft4_type_validator, minimal_swagger_spec):
+    prop_schema = {'x-nullable': False, 'type': 'string'}
+    args = (None, prop_schema['type'], None, prop_schema)
+    list(type_validator(minimal_swagger_spec, *args))
```
sjaensch: I'd use keyword arguments here as well, just like you did in the test above. It's more readable.

tests/unmarshal/unmarshal_nullable_test.py
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    try:
+        validate_schema_object(empty_swagger_spec, content_spec, value)
+        result = unmarshal_schema_object(
+            empty_swagger_spec, content_spec, value)
+    except ValidationError as e:
+        result = e
+
+    if nullable is False and not value:
```
sjaensch: `not value is None` is more correct here. Which reminds me, testing with an empty string (or a bool with the value False) might not be a bad idea.

You could split out the case that is supposed to fail into a separate test function so that you don't have to have this logic here at all, but I'm not sure if that actually improves things.

tests/unmarshal/unmarshal_nullable_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if value == {} and required is True:
+        assert result.message == "'x' is a required property"
+    elif value == {} and required is False:
+        # Unmarshal re-introduces missing properties with None which is ok.
+        assert result == {'x': None}
+    elif nullable is False and value['x'] is None:
+        assert result.message == "None is not of type 'string'"
+    else:
+        result == value
```
sjaensch: Having such a long `if/elif` when doing test result assertion is a good indicator that your test tries to do too many things at once. In the failure case you could then use `pytest.raises` to check that the ValidationError was thrown and use mark.parametrize for the expected error message.

In such a case where the expected outcome is dependent on several parameters you could use the singular version of mark.parametrize (i.e. only one decorator that specifies all parameter combinations explicitly) together with the expected outcome.

tests/swagger20_validator/type_validator_test.py
andreashug: This is because `assert_called_once_with()` expects the arguments without keywords. The existing tests in this file do it this way so I aligned with them ;-)

tests/unmarshal/unmarshal_nullable_test.py
andreashug: You are totally right. When I started to write these integration tests, I actually didn't think that they should go into repo (still not sure about that) because they only clarify what happens in which scenario. But anyway, here's the refactoring =)

andreashug:
Regarding `marshal_object()`: The `ValidationError` is raised when `x-allownull` is `True`, the value is required the value is set to `None`. See https://dpaste.de/D0ax

sjaensch:
This is great! Could you please create another paste? I didn't manage to open it in time. 😁

andreashug:
Sure: http://pastebin.com/U3xeA48R  (Seems like there is a bug in expiration settings on dpaste.de)

sjaensch:
Thanks again @andreashug! While I think we could even ship this without fixing the edge case about a required x-nullable parameter, I think the fix is not that difficult:

```
diff --git a/bravado_core/marshal.py b/bravado_core/marshal.py
index 536e627..8a3771a 100644
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -125,13 +125,14 @@ def marshal_object(swagger_spec, object_spec, object_value):
     result = {}

-        # Values cannot be None - skip them entirely!
-        if v is None:
-            continue
-
             swagger_spec, deref(object_spec), object_value, k)

+        # Values cannot be None unless x-nullable is set
+        is_nullable = prop_spec and prop_spec.get('x-nullable', False)
+        if v is None and not is_nullable:
+            continue
+
```

With it the test you wrote (which is in your dpaste) passes. We might want to handle the other cases (nullable primitives etc.) as well, but honestly I think it's not that important, since the use case for x-nullable are model properties in responses. You're welcome to incorporate this small change as well as the test in the branch though.

I'm going to find some time to test this with bravado and make sure it does what we want. 😊

sjaensch:
An manual test using bravado to query an endpoint that returns a model with an x-nullable field worked perfectly! I'm going to merge this in and work on a second pull request that will contain my above change, a version bump and a big thank you to @andreashug in the changelog. 😊

andreashug:
Yay, thanks  \o/

Issue #70, opened 2016-02-19
casting of boolean values doesn't work


watterso:
See the below code excerpt from `bravado_core/param.py` for proof.
`param_value` comes in as a string, using `bool()` as a casting function would result in the following behavior:
* `bool('non-empty-string') == True`
* `bool('') == False` (doesn't happen because None is returned if `param_value == ''`)
```python
CAST_TYPE_TO_FUNC = {
    'integer': int,
    'number': float,
    'boolean': bool
}
def cast_request_param(param_type, param_name, param_value):
    ....
    if param_value is None:
        return None

    if param_type in CAST_TYPE_TO_FUNC and param_value == '':
        return None

    try:
        return CAST_TYPE_TO_FUNC.get(param_type, lambda x: x)(param_value)
    except ValueError:
        log.warn("Failed to cast %s value of %s to %s",
                 param_name, param_value, param_type)
        # Ignore type error, let jsonschema validation handle incorrect types
        return param_value
```

PR #71, opened 2016-02-19
fix parsing boolean values

watterso:
Fixes issue #70 

I thought accepting case-insensitive values of 'true' and 'false' would be appropriate, thoughts?

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -186,7 +186,7 @@ def unmarshal_param(param, request):
     return value
 
 
-def _booleans_are_true_false_or_0_1_strings(value):
+def string_to_boolean(value):
     lowercase_value = value.lower()
     true = ['true', '1']
     false = ['false', '0']
@@ -204,7 +204,7 @@ CAST_TYPE_TO_FUNC = {
     # cast them to the right type
     'integer': int,
     'number': float,
-    'boolean': _booleans_are_true_false_or_0_1_strings
+    'boolean': string_to_boolean
 }
 
 
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -186,11 +186,13 @@ def unmarshal_param(param, request):
     return value
 
 
-def _booleans_are_true_false_strings(value):
+def _booleans_are_true_false_or_0_1_strings(value):
     lowercase_value = value.lower()
-    if lowercase_value == 'true':
+    true = ['true', '1']
+    false = ['false', '0']
+    if lowercase_value in true:
         return True
-    elif lowercase_value == 'false':
+    elif lowercase_value in false:
         return False
     else:
         # raise ValueError so casting failure will be logged
@@ -202,7 +204,7 @@ CAST_TYPE_TO_FUNC = {
     # cast them to the right type
     'integer': int,
     'number': float,
-    'boolean': _booleans_are_true_false_strings
+    'boolean': _booleans_are_true_false_or_0_1_strings
 }
 
 
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -192,11 +192,10 @@ def string_to_boolean(value):
     false = ['false', '0']
     if lowercase_value in true:
         return True
-    elif lowercase_value in false:
+    if lowercase_value in false:
         return False
-    else:
-        # raise ValueError so casting failure will be logged
-        raise ValueError()
+    # Failed casts raise a ValueError
+    raise ValueError()
 
 
 CAST_TYPE_TO_FUNC = {
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -188,11 +188,10 @@ def unmarshal_param(param, request):
 
 def _booleans_are_true_false_strings(value):
     lowercase_value = value.lower()
-    if lowercase_value in ['true', 'false']:
-        if lowercase_value == 'true':
-            return True
-        else:
-            return False
+    if lowercase_value == 'true':
+        return True
+    elif lowercase_value == 'false':
+        return False
     else:
         # raise ValueError so casting failure will be logged
         raise ValueError()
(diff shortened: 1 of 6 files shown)


watterso:
ping @analogue because you responded to my last PR :)

analogue:
Does it make sense to also support 0 and 1?

watterso:
Sounds good, I'll add that in real quick

watterso:
@analogue :ship: ? :smiley:

bravado_core/param.py
```diff
@@ -186,10 +186,25 @@ def unmarshal_param(param, request):
     return value
 
 
+def _booleans_are_true_false_or_0_1_strings(value):
```
analogue: Would be nice to stay consistent with the rest of the codebase here - punt on using leading underscore.

watterso: Seems reasonable!

bravado_core/param.py
```diff
@@ -186,10 +186,25 @@ def unmarshal_param(param, request):
...
+    lowercase_value = value.lower()
+    true = ['true', '1']
+    false = ['false', '0']
+    if lowercase_value in true:
+        return True
+    elif lowercase_value in false:
```
analogue: s/elif/if

bravado_core/param.py
```diff
@@ -186,10 +186,25 @@ def unmarshal_param(param, request):
...
+        return False
+    else:
```
analogue: else can be deleted

tests/param/cast_request_param_test.py
```diff
@@ -11,9 +17,29 @@ def test_empty_string_becomes_none_for_type_integer():
...
+    assert not cast_request_param('boolean', 'is_open', 'false')
+    assert not cast_request_param('boolean', 'is_open', 'faLSe')
+    assert not cast_request_param('boolean', 'is_open', '0')
+
+
+@patch('bravado_core.param.log')
+def test_boolean_returns_unchanged_string_for_non_bool_strings(mock_logger):
+    assert cast_request_param('boolean', 'is_open', 'PIZZA') == 'PIZZA'
+    assert not cast_request_param('boolean', 'is_open', 'PIZZA') is True
+    assert not cast_request_param('boolean', 'is_open', 'PIZZA') is False
```
analogue: Not sure I understand the purpose of the last two assertions.  The first assertion covers the test case.

tests/param/cast_request_param_test.py
```diff
@@ -3,6 +3,12 @@
 from bravado_core.param import cast_request_param
 
 
+@patch('bravado_core.param.log')
+def test_logs_cast_failure(mock_logger):
+    cast_request_param('integer', 'gimme_int', 'not_int')
```
analogue: I'd prefer unit tests on the `_booleans_are_true_false_or_0_1_strings(...)` function itself, but this is fine.

tests/param/cast_request_param_test.py
```diff
@@ -11,9 +17,29 @@ def test_empty_string_becomes_none_for_type_integer():
...
+
+@patch('bravado_core.param.log')
+def test_boolean_returns_unchanged_string_for_non_bool_strings(mock_logger):
+    assert cast_request_param('boolean', 'is_open', 'PIZZA') == 'PIZZA'
+    assert not cast_request_param('boolean', 'is_open', 'PIZZA') is True
+    assert not cast_request_param('boolean', 'is_open', 'PIZZA') is False
+
+
+@patch('bravado_core.param.log')
+def test_boolean_cast_failure_is_logged(mock_logger):
```
analogue: This is a variation of the test on line 7.  Not hurting anything, but drives home the point of testing the function instead of testing a function that calls another function (hence resulting in duplication).

tests/param/cast_request_param_test.py
watterso: Looks like cruft that didn't get removed when I was cleaning up my tests, will remove.

tests/param/cast_request_param_test.py
watterso: This tests that cast failures in general are logged, a test that was previously missing.
Further down I'll change the tests to be calling the `string_to_boolean` (formerly `_booleans_......`) function directly.

tests/param/cast_request_param_test.py
watterso: I'll have this test call the function directly and check that a `ValueError` is raised rather than checking by proxy via the logging behavior.

bravado_core/param.py
watterso: ack

watterso:
@analogue please excuse the messy extra commits xD
What do you think now?

watterso:
Ughhh i messed up changelog and maybe other files because of bad merge, i'll fix that later, please look at the rest though.

tests/param/cast_request_param_test.py
```diff
@@ -1,42 +1,63 @@
 from mock import patch
+import pytest
 
 from bravado_core.param import cast_request_param
+from bravado_core.param import string_to_boolean
 
 
-def test_integer():
-    assert 34 == cast_request_param('integer', 'biz_id', '34')
+class TestGeneralCast:
```
analogue: Please, no test classes for consistency - look at all the other tests in this codebase. The general pattern is a `<function>_test.py` file per function under test.  This has so many benefits over putting the tests for a single module in a single file - no more 1000+ line unit test `.py` files :-)

- Add/update tests for `cast_request_param()` to `cast_request_param_test.py`
- Create `string_to_bool_test.py` and add unit tests as needed

watterso: :( Okay I'll do this, I still like test classes, because you can group tests based on a theme, like behavior on an empty string vs general behavior. But I'll remove them for consistency.

analogue:
:+1:

watterso:
:D

Issue #42, opened 2015-08-12
Support yaml format for swagger spec as well


prat0318:
Currently we support swagger spec in `json` format. It will be nice to have `yaml` support as well. I think it precedes Yelp/bravado#168

analogue:
Feature added in 4.1.0

PR #73, opened 2016-02-25
Add yaml support for swagger specs

djeebus:
This implements #42, hopefully =) I ended up having to override `jsonschema`'s `file` protocol handler in order to sniff them properly. I also added a dependency on the `requests` library for testing.

The test got a little bit long, but I ended up implementing a tiny web server in order to verify that the client works. It got weird, but I think it works well.

Thanks for the work on the project! I found this through `pyramid_swagger`, I'm trying to implement yaml on their side, which led me here.


djeebus:
Ah, wasn't able to test py2.6 on my machine. I'll fix this, sorry.

bravado_core/spec.py
```diff
@@ -213,11 +239,24 @@ def download(uri):
...
+        fp = urlopen(uri)
+        if is_yaml(uri):
+            return yaml.load(fp)
+        else:
+            return json.loads(fp.read().decode("utf-8"))
 
     return {
         'http': download,
         'https': download,
+        'file': read_file,
```
analogue: It wasn't obvious to me you were overriding jsonschema's built in `file` handler (even though you mention it in the PR description). Can you add a comment to point this out and why?

bravado_core/spec.py
```diff
@@ -213,11 +239,24 @@ def download(uri):
             'method': 'GET',
             'url': uri,
         }
-        return http_client.request(request_params).result().json()
+        response = http_client.request(request_params).result()
+        content_type = response.headers.get('content-type', '').lower()
```
analogue: I'm guessing `headers` has case-insensitivity baked in.

setup.py
```diff
@@ -35,5 +35,6 @@
         "six",
         "simplejson",
         "swagger-spec-validator>=2.0.1",
+        "pyyaml",
```
analogue: Please keep sorted :smile:

test-data/2.0/yaml/pet.yaml
```diff
@@ -0,0 +1,24 @@
+---
+  definitions:
```
analogue: Is the extra indentation in all but `fake.yiml` intentional?

tests/spec/from_dict_test.py
```diff
@@ -83,6 +84,38 @@ def test_ref_to_external_path_with_ref_to_local_model():
...
+
+    with open(swagger_yaml_path) as f:
+        swagger_yaml_content = yaml.load(f)
+
+    swagger_yaml_url = urlparse.urljoin('file:', swagger_yaml_path)
+    spec = Spec.from_dict(swagger_yaml_content, swagger_yaml_url)
+    assert 'Pet' in spec.definitions
+
+
+def test_file_requests():
```
analogue: I don't think this test belongs in `from_dict_test.py`.  The `test_yaml_files()` seems to cover this behavior, though.

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,107 @@
...
+            from http.server import HTTPServer, SimpleHTTPRequestHandler
+        except ImportError:
+            from BaseHTTPServer import HTTPServer
+            from SimpleHTTPServer import SimpleHTTPRequestHandler
+
+        ext_map = SimpleHTTPRequestHandler.extensions_map
+        for file_extension in ('.yaml', '.yml', '.yiml'):
+            ext_map[file_extension] = 'application/x-yaml'
+
+        listen_address = ('127.0.0.1', 19283)
```
analogue: This is just asking to fail :disappointed:

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,107 @@
+import json
+import os
+import requests
+import unittest
+import yaml
+
+
+class TestHttpClients(unittest.TestCase):
```
analogue: Following this test made my head hurt :cry:  How about something simpler that just exercises the `http` handler returned by `build_http_handlers()` and mocks out `http_client`?  No server necessary.

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py26, py27, py34, flake8
 [testenv]
 deps =
     -rrequirements-dev.txt
+    requests
```
analogue: Don't think you'll end up needing this, but it can be added to the actual `requirements-dev.txt` file.

test-data/2.0/yaml/pet.yaml
djeebus: Whoops, no, i just ran some json files through a translator and it came out indented. I can clean that up.

setup.py
djeebus: Will do! Thanks for wanting to keep things clean =)

tests/spec/build_http_handlers_test.py
djeebus: Haha, ok, will do.

bravado_core/spec.py
djeebus: Do you want a comment added in `bravado_core/spec.py`?

bravado_core/spec.py
djeebus: Yup:

https://github.com/kennethreitz/requests/blob/master/requests/models.py#L550

djeebus:
@analogue should look a bit better now.

bravado_core/spec.py
analogue: Thanks for chasing that down.

bravado_core/spec.py
analogue: Yup, that would be great.

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,106 @@
...
+    mock_response = mock.Mock()
+    mock_response.content = StringIO(content)
+    mock_response.json.side_effect = lambda *args, **kwargs: json.loads(content)
+
+    mock_http_client = mock.Mock()
+    result_func = mock_http_client.request.return_value.result
+
+    result_func.return_value = mock_response
+
+    from bravado_core.spec import build_http_handlers
```
analogue: Can you move this to the top of the file with the rest of the imports?

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,106 @@
...
+    handlers, response = _build_http_client(test_json)
+    response.headers = {'content-type': 'text/plain'}
+
+    http_handler = handlers['http']
+    result = http_handler('https://test.test/json/test.json')
+
+    assert result == test_dict
+
+
+def test_http_handler_equals_https_handler():
```
analogue: Not sure this test serves any useful purpose.

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,106 @@
...
+except ImportError:
+    import mock
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+
+def _run_test(self, path, loader):
```
analogue: Doesn't look like this method is referenced anywhere.

tests/spec/build_http_handlers_test.py
djeebus: It was just to make sure that the, in lieu of duplicating tests to prove that both `http` and `https` protocol handlers work, we just make sure that they're both the same. This would be a canary in a coal mine: if the protocol handlers ever differ, this test would fail, and tests could be written.

That was just the thought behind it; I can drop it if you like.

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,106 @@
...
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
```
analogue: Why are the above two try/except's necessary? `mock` is in `requirements-dev.txt`.

requirements-dev.txt
```diff
@@ -2,3 +2,4 @@
 mock<1.1.0
 pre-commit
 pytest
+requests
```
analogue: I don't see `requests` imported anywhere. Can safely remove?

tests/spec/build_http_handlers_test.py
djeebus: Whoops! Zombie code ;)

test-data/2.0/yaml/toys.yaml
```diff
@@ -0,0 +1,24 @@
+---
+definitions:
+  Toy:
```
analogue: Don't see this file used any longer. Can safely remove?

tests/spec/build_http_handlers_test.py
```diff
@@ -0,0 +1,106 @@
...
+def test_http_handler_equals_https_handler():
+    """
+    The HTTPServer class doesn't support https. Instead of
+    testing https directly, we'll just make sure that the two handlers
+    are identical.
+    :return:
+    """
+
+    from bravado_core.spec import build_http_handlers
```
analogue: delete

tests/spec/build_http_handlers_test.py
analogue: Ah, `pytest` has just what you need in this case.  https://pytest.org/latest/example/parametrize.html

Just parameterize http and https to the existing tests :smile:

tests/spec/build_http_handlers_test.py
djeebus: `mock` was added to the standard library under the `unittest` module in python 3.3. I was attempting to use that one before reverting to the external module. I can just rely on the external module if you like.

`StringIO` moved in py3 to the `io` library. These imports are to ensure py2/py3 compatibility.

analogue: Will this work for `StringIO`? http://python-future.org/compatible_idioms.html#stringio

requirements-dev.txt
djeebus: Argh, this is what I get for trying to fix a PR past midnight. Thanks for being patient :)

bravado_core/spec.py
djeebus: np!

tests/spec/build_http_handlers_test.py
djeebus: Interesting, didn't know that they backported the reference to `io`. I'll shorten both of these.

On a side note, this PR is going to need a rebase before it gets merged =)

analogue:
Looking good! Thanks for contributing.

djeebus:
Not a problem, thanks for the merge!

Issue #66, opened 2016-01-15
Validation fails for arrays of non-strings  when using the "multi" collection format


JoelO:
Examples of failed validation:

When queried with `value=2`
```
<html>
 <head>
  <title>400 Bad Request</title>
 </head>
 <body>
  <h1>400 Bad Request</h1>
  The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>
u'2' is not of type 'array'

Failed validating 'type' in schema:
    {'collectionFormat': 'multi',
     'description': 'Value',
     'in': 'query',
     'items': {'format': 'int32', 'type': 'integer'},
     'name': 'value',
     'required': False,
     'type': 'array'}

On instance:
    u'2'


 </body>
</html>
```

and when queried with `value=2&value=1`
```
<html>
 <head>
  <title>400 Bad Request</title>
 </head>
 <body>
  <h1>400 Bad Request</h1>
  The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>
u'2' is not of type 'integer'

Failed validating 'type' in schema['items']:
    {'format': 'int32', 'type': 'integer'}

On instance[0]:
    u'2'

 </body>
</html>
```

analogue:
Fixed in 4.1.0

PR #75, opened 2016-02-26
Fix validation on non string multi collection

JoelO:
Fixes issue #66.

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -303,18 +303,21 @@ def unmarshal_collection_format(swagger_spec, param_spec, value):
 
     if collection_format == 'multi':
         # http client lib should have already unmarshaled to an array
-        return value
+        value_array = value
 
-    if value is None and not schema.is_required(swagger_spec, param_spec):
+    elif value is None and not schema.is_required(swagger_spec, param_spec):
         # Just pass through an optional array that has no value
         return None
 
-    sep = COLLECTION_FORMATS[collection_format]
+    else:
+        sep = COLLECTION_FORMATS[collection_format]
+        value_array = value.split(sep)
+
     items_spec = param_spec['items']
     items_type = deref(items_spec).get('type')
     param_name = param_spec['name']
 
     return [
         cast_request_param(items_type, param_name, item)
-        for item in value.split(sep)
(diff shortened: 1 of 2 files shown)


bravado_core/param.py
```diff
@@ -317,18 +317,21 @@ def unmarshal_collection_format(swagger_spec, param_spec, value):
 
     if collection_format == 'multi':
         # http client lib should have already unmarshaled to an array
-        return value
+        value_array = value
```
analogue: Won't this fail for the case where the cardinality of the value is one? Unless I'm missing something, the http client lib has no idea that the single item should be wrapped in an array.  
For example: 
- `http://foo.com/bar?ids=1`, the `value` is `"1"`
- `http://foo.com/bar?ids=1&ids=2`, the `value` is `["1", "2"]`

tests/param/unmarshal_param_test.py
```diff
@@ -98,6 +113,26 @@ def test_optional_query_array_with_default(
...
+    param = Param(empty_swagger_spec,
+                  Mock(spec=Operation),
+                  int_array_param_spec)
+    request = Mock(
+        spec=IncomingRequest,
+        query={'numbers': ['1', '2', '3']})
+    assert [1, 2, 3] == unmarshal_param(param, request)
+
+
+def test_query_int_singleton_array(empty_swagger_spec, int_array_param_spec):
```
analogue: Looks like this can be merged into a single test (`test_query_int_array`) by using pytest's support for parameterization. 

https://pytest.org/latest/parametrize.html

bravado_core/param.py
```diff
@@ -315,20 +315,23 @@ def unmarshal_collection_format(swagger_spec, param_spec, value):
...
-
     if value is None and not schema.is_required(swagger_spec, param_spec):
         # Just pass through an optional array that has no value
         return None
 
-    sep = COLLECTION_FORMATS[collection_format]
+    if collection_format == 'multi':
+        # http client lib should have already unmarshaled to an array
+        value_array = value if isinstance(value, list) else [value]
+
```
analogue: delete blank line?

PR #76, opened 2016-03-01
Sanitize operationId

cyprieng:
See Yelp/bravado#189

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+import re
 
 from bravado_core.param import Param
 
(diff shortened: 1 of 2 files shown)


bravado_core/operation.py
```diff
@@ -88,12 +88,17 @@ def operation_id(self):
...
-                self._operation_id = (self.http_method + '_' + self.path_name)\
-                    .replace('/', '_')\
-                    .replace('{', '_')\
-                    .replace('}', '_')\
-                    .replace('__', '_')\
-                    .strip('_')
+                self._operation_id = (self.http_method + '_' + self.path_name)
+
+        # Sanitize operation ID
+        self._operation_id = self._operation_id.replace(' ', '_')\
```
analogue: No need to sanitize more than once.

analogue: Also, probably a good time to turn this into a more manageable regex:
```
r = re.compile('[^A-Za-z0-9_]|_+')
self._operation_id = r.sub('_', self._operation_id).strip('_')
```

tests/operation/operation_id_test.py
```diff
@@ -21,3 +21,10 @@ def test_returns_generated_operation_id_with_path_parameters():
     operation_spec = {}
     operation = Operation(spec, '/pet/{petId}', 'get', operation_spec)
     assert 'get_pet_petId' == operation.operation_id
+
+
+def test_returns_generated_operation_id_when_badly_formatted():
```
analogue: `test_returns_sanitized_operation_id_when_using_illegal_chars()` sounds a bit more accurate.

Issue #78, opened 2016-03-07
x-scope annotations on server side shouldn't be sent to the client


analogue:
See internal ticket GSI-252 for the details.

Basically, x-scope annotations that reference `file://` type urls are being sent to the client which then tries to use the `file://` type urls to resolve $refs.  The x-scope annotation should not be part of the swagger.json sent to the client.

The error surfaces on the client side in an exception that closely resembles:

```RefResolutionError: <urlopen error [Errno 2] No such file or directory: '/code/api_docs/swagger.json'>```

analogue:
Fixed in 4.2.0

PR #79, opened 2016-03-09
Make spec_dict available without internal x-scope metadata

analogue:
Fix for #78 

Server-side libraries that use bravado-core (such as pyramid-swagger) now have access to the Swagger spec in dict form without the internally attached x-scope metadata.

```python
spec = Spec.from_dict(spec_dict)
# spec_dict suitable for serving up to Swagger clients
print spec.client_spec_dict
```
Given the options available, this bubbled up as the most straight-forward way to maintain backward compatibility, retain the mutability necessary to tag models, and provide server-side libraries access to the stripping behavior for subsequent remote $ref requests via ``bravado_core.spec.strip_xscope``.

Tested with example_happyhour (a single line change was necessary to pyramid_swagger - will submit PR after this is merged and released).

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import copy
 import functools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


prat0318:
lgtm

bravado_core/spec.py
```diff
@@ -70,6 +71,9 @@ def __init__(self, spec_dict, origin_url=None, http_client=None,
         self.api_url = None
         self.config = dict(CONFIG_DEFAULTS, **(config or {}))
 
+        # Cached copy of spec_dict with x-scope metadata removed. See @property
```
sjaensch: Did you mean "see client_spec_dict()"?

sjaensch:
lg2m2

bravado_core/spec.py
analogue: :+1:

Issue #77, opened 2016-03-01
Unmarshaling request with non-required enum fails


jacobstr:
`unmarshal_request` delegates to `bravado_core.validate.validate_primitive` which fails validation when `None` is passed in for a non-required parameter.

```
>           raise error
E           ValidationError: None is not one of ['inky', 'dinky', 'doo']
E
E           Failed validating 'enum' in schema:
E               {'enum': ['inky', 'dinky', 'doo'], 'type': 'string'}
E
E           On instance:
E               None
```

I've pushed a branch with a test that captures this:

https://github.com/jacobstr/bravado-core/tree/validate-non-required-enum

analogue:
Swagger doesn't support passing null as a value.  Hence, if the parameter is not required, the parameter is not present in the first place.

jacobstr:
In that case, should `unmarshal_request` be iterating over non-required parameters that haven't been provided?

analogue:
Yes, how else would it know to use the `default` value if one is provided in the schema in the absence of the non-required parameter?

jacobstr:
Perhaps I'm simply using the API incorrectly - the ultimate intent is to be able to use `unmarshal_request` against an endpoint that has an optional string parameter in the query.

```
spec = Spec.from_dict(spec)
op = spec.resources['things'].operations['ListThings']
data = unmarshal_request(request, op)
```

Suppose that ListThings has a `category` enum. If the request.query dictionary is an empty dictionary, `unmarshal_request` will fail.

analogue:
Ok, I think I see what you mean. Looks like the [enum validator](https://github.com/Yelp/bravado-core/blob/master/bravado_core/swagger20_validator.py#L58) may need some additional code to handle this case.

analogue:
Fixed in 4.2.1

PR #80, opened 2016-03-15
Fix request params that are optional enums

analogue:
Fix for #77


analogue:
@sjaensch  or @prat0318

sjaensch:
I already took a look at it. Looks sane to me, but I'm not yet too familiar with the internals of bravado-core. 👍from me.

PR #83, opened 2016-03-21
Use x-scope annotations when resolving refs during request/response validation

analogue:
Fix for #82

Unit tests have been added.

PR #84, opened 2016-03-23
Use x-scope annotations when resolving refs during request/response validation

### LINKED ISSUE #82: Fix ref resolution during response validation
Ref resolution during response validation sometimes fails because of where the validation starts. For refs to resolve 100% of the time, the scope maintained by jsonschema's RefResolver has to be rooted at `/` in swagger.json, not at the response_spec fragment that is currently passed to the `validate_response(..)` function.  

Luckily, the full scope is available as `x-scope` annotations in the spec so the `ref` validator just needs to be customized to use the `x-scope` annotation if it is available.

This is quite difficult to explain, so an example is probably useful to compare working and broken scopes for the RefResolver for an internal use case:

Broken RefResolver scope:
```
['file:///nail/home/spatel/pg/services/biz_app/api_docs/swagger.json']
```

Working RefResolver scope:
```
[
'file:///nail/home/spatel/pg/services/biz_app/api_docs/swagger.json',
'file:///nail/home/spatel/pg/services/biz_app/api_docs/tags/account.json#/paths/account.photo.v1',
'file:///nail/home/spatel/pg/services/biz_app/api_docs/tags/account.json#/definitions/AccountPhotoResponse'
]
```

sjaensch:
Fixes #82, closes #83.

Created a second PR based on the fix by @analogue. The only change is fixing and enabling the second test. I verified internally that our service works with this fix. Your change lg2m, thumbs up from me for merging this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@ Changelog
 4.2.1 (2016-XX-XX)
 ------------------
 - Fix optional enums in request params - Issue #77
+- Fix resolving refs during validation - Issue #82
 
 4.2.0 (2016-03-10)
 ------------------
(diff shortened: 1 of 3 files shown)


analogue:
Awesome! :+1:

PR #101, opened 2016-07-25
Accept tuples when validating list-like params

mjbryant:


prat0318:
what is the use-case of such a change. imo tuples are mostly fixed sized and won't be an ideal fit for "lists", but yeah, i may be missing something.

asottile:
No request parameter should be mutable.  tuples actually make *more* sense than lists

mjbryant:
@prat0318 as it stands, you're not able to make a Bravado service call with a tuple instead of a list. E.g. if the `make_service_call` resource takes a single array-type parameter, `user_ids`, this will fail:

```
client = SwaggerClient.from_url(...)
result = client.tag.make_service_call(user_ids=(1, 2, 3))
```

will fail with `bravado_core.exception.SwaggerMappingError: Expected list like type for <class 'tuple'>`. This seems ludicrous.

bravado_core/schema.py
```diff
@@ -57,7 +57,7 @@ def is_list_like(spec):
     :rtype: boolean
     """
```
analogue: Can you fix the docs while you're in there? :+1:

bravado_core/schema.py
```diff
@@ -57,7 +57,7 @@ def is_list_like(spec):
...
     # TODO: check magic method instead
```
analogue: Delete this too

mjbryant:
@analogue good?

analogue:
:ship:

mjbryant:
@analogue could you merge this in? Trying to use this version in a CR so I don't have to use bad hax.

PR #103, opened 2016-08-01
Refactor nullable-related tests, add support for marshalling

sjaensch:
I've refactored the nullable-related tests by @andreashug to better reflect the structure of the other tests: I've split them up into unmarshalling and validation tests and added them the the existing files. I've also added support for marshalling along with some tests as well as a bit of documentation.


tests/unmarshal/unmarshal_primitive_test.py
```diff
@@ -72,3 +72,15 @@ def test_ref(minimal_swagger_dict):
     special_integer_spec = {'$ref': '#/definitions/SpecialInteger'}
     swagger_spec = Spec.from_dict(minimal_swagger_dict)
     assert 10 == unmarshal_primitive(swagger_spec, special_integer_spec, 10)
+
+
+@pytest.mark.parametrize(['nullable', 'value'],
+                         [(False, 'x'), (True, 'x'), (True, None)])
+def test_nullable(minimal_swagger_spec, value, nullable):
+    """Test scenarios in which validation should pass: (1), (3), (4)"""
```
laucia: that docstring doesn't means anything now, you need to explain the cases tested explicitely

sjaensch: Thanks for catching that! The validation is in the other tests now, I removed the docstring.

docs/source/models.rst
```diff
@@ -78,6 +78,25 @@ always disable it if necessary.
     swagger_dict = {..}
     spec = Spec.from_dict(swagger_dict, config={'use_models': False})
 
+Allowing null values for properties
+-----------------------------------
+Typically, bravado-core will complain during validation if it encounters fields with ``null`` values.
+This can be problematic, especially when you're adding Swagger support to pre-existing
+APIs. In that case, declare your model properties as ``x-nullable``:
```
laucia: I was wondering if we should me more explicit about the fact that this is a Vendor Extension and not part of the regular swagger-spec ?

tests/unmarshal/unmarshal_object_test.py
```diff
@@ -236,3 +236,44 @@ def test_recursive_ref_with_depth_n(recursive_swagger_spec):
...
+    assert result == value
+
+
+@pytest.mark.parametrize('nullable', [True, False])
+def test_nullable_no_value(empty_swagger_spec, nullable):
+    content_spec = nullable_spec_factory(False, nullable=nullable)
+    value = {}
+
+    result = unmarshal_object(empty_swagger_spec, content_spec, value)
+    assert result == {'x': None}  # Missing parameters are re-introduced
```
laucia: I am probably wrong but I don't get why we would unmarshall to this when nullable is `False` and required is false too. 
That look like the old swaggerpy behavior of adding `None` on not present key.

sjaensch: swaggerpy re-added the field with the type's default value, not ``None``. It looks like the current behavior is desired, check out this comment in [unmarshal.py](https://github.com/Yelp/bravado-core/blob/master/bravado_core/unmarshal.py#L124).

docs/source/models.rst
sjaensch: Done, what do you think?

laucia:
lg2m 🚢

PR #107, opened 2016-08-05
Check if parameter value is a bool-type before assuming it's a string

nickdirienzo:
Hey there!

I use pyramid_swagger, which relies on bravado-core. pyramid_swagger converts primitives when it can, which happens before it ends up calling bravado-core's `unmarshal_request`.

An issue I ran into today is essentially the test I've included with this change. I have a query parameter of a boolean type and by the time pyramid_swagger sends it to bravado_core for unmarshalling, the boolean is already a Python-bool and not a string-bool (i.e. 'true'), which causes `string_to_boolean` to raise a `ValueError`.

This change checks to see if a value passed into `string_to_boolean` is of type `bool` before treating it like a string.

I think it makes sense to fix this issue in this project so people who build on top of this can send Python bools if they want (which seems reasonable?). However, if this is not the right place to fix this problem, and instead I should be fixing this in pyramid_swagger, let me know.

Thanks.

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -198,8 +198,11 @@ def unmarshal_param(param, request):
 
 
 def string_to_boolean(value):
-    """Coerce the provided boolean-typed parameter value into its Python
-    boolean value if it's a string or return the value as-is if already casted.
+    """Coerce the provided value into its Python boolean value if it's a string
+    or return the value as-is if already casted.
+
+    :param value: the value of a Swagger parameter with a boolean type
+    :type value: usually string, but sometimes a bool
     """
     if isinstance(value, bool):
         return value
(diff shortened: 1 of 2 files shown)


nickdirienzo:
@sjaensch would you be the right person to take a look at this? Thanks!

sjaensch:
@nickdirienzo absolutely! I see no harm in doing this in bravado-core, it does make the code more robust. Going to merge it in, thanks for the pull request!

nickdirienzo:
@sjaensch awesome. Thanks for merging it in! What's the release cycle for bravado-core? Do you know when a patch version bump will come out containing this fix? It'd be great if I didn't have to point our bravado-core dependency to a commit hash and instead could point it to 4.3.1. :)

sjaensch:
@nickdirienzo I just released 4.3.1, I hope that helps. 😊

nickdirienzo:
@sjaensch thank you! 😊 (and thanks for maintaining this awesome project.)

Issue #108, opened 2016-08-09
Error during unmarshal_object for x-nullable=true property


zroadhouse-rmn:
I believe unmarshal_object code path is missing support for x-nullable. I get the following traceback when unmarshaling a property which a specification like:

~~~
definitions:
  SomeType:
    properties:
      extra_data:
        type: object
        x-nullable: true
~~~

Here's the stack trace:

~~~
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado/http_future.py", line 77, in result
    self.response_callbacks)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado/http_future.py", line 115, in unmarshal_response
    operation)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/response.py", line 110, in unmarshal_response
    op.swagger_spec, content_spec, content_value)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 52, in unmarshal_schema_object
    return unmarshal_object(swagger_spec, schema_object_spec, value)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 124, in unmarshal_object
    result[k] = unmarshal_schema_object(swagger_spec, prop_spec, v)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 42, in unmarshal_schema_object
    return unmarshal_array(swagger_spec, schema_object_spec, value)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 100, in unmarshal_array
    for item in array_value
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 52, in unmarshal_schema_object
    return unmarshal_object(swagger_spec, schema_object_spec, value)
  File "/Users/zroadhouse/.virtualenvs/dartclient-bravado/lib/python2.7/site-packages/bravado_core/unmarshal.py", line 124, in unmarshal_object
[...]

PR #109, opened 2016-08-09
Fix to marshalling to better handle x-nullable and required properties

zroadhouse-rmn:
This is the proposed fix for https://github.com/Yelp/bravado-core/issues/108

I updated unmarshal to better handle x-nullable properties and added additional test cases to cover the changes.

There are a number of remaining issues which I would prefer to leave for someone else to address:

- Unmarshalling of required non-nullable arrays works when it probably shouldn't (existing behavior)
- I had to make a copy of the dereferenced property specification dictionary to add in the x-nullable flag if present. My experience with serde in the past is that I often needed to pass the property specification in addition to the referenced type information. I chose not to do that here for expediency. The extra dictionary copy could be optimized by refactoring the code to do so.

### DIFF
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -5,7 +5,7 @@ from bravado_core.exception import SwaggerMappingError
 from bravado_core.model import is_model, MODEL_MARKER
 from bravado_core.schema import is_dict_like
 from bravado_core.schema import is_list_like
-from bravado_core.schema import is_prop_nullable
+from bravado_core.schema import handle_null_value
 from bravado_core.schema import SWAGGER_PRIMITIVES
 from bravado_core.schema import get_spec_for_prop
 
@@ -74,11 +74,7 @@ def marshal_primitive(swagger_spec, primitive_spec, value):
         value = schema.get_default(swagger_spec, primitive_spec)
 
     if value is None:
-        if is_prop_nullable(swagger_spec, primitive_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(primitive_spec))
+        return handle_null_value(swagger_spec, primitive_spec)
 
     if not default_used:
         value = formatter.to_wire(swagger_spec, primitive_spec, value)
@@ -96,11 +92,7 @@ def marshal_array(swagger_spec, array_spec, array_value):
     :raises: SwaggerMappingError
     """
     if array_value is None:
-        if is_prop_nullable(swagger_spec, array_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(array_spec))
+        return handle_null_value(swagger_spec, array_spec)
 
     if not is_list_like(array_value):
         raise SwaggerMappingError('Expected list like type for {0}: {1}'
@@ -130,11 +122,7 @@ def marshal_object(swagger_spec, object_spec, object_value):
     deref = swagger_spec.deref
 
     if object_value is None:
-        if is_prop_nullable(swagger_spec, object_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(object_spec))
+        return handle_null_value(swagger_spec, object_spec)
 
     if not is_dict_like(object_value):
         raise SwaggerMappingError('Expected dict like type for {0}:{1}'.format(
@@ -142,11 +130,6 @@ def marshal_object(swagger_spec, object_spec, object_value):
 
     object_spec = deref(object_spec)
     required_fields = object_spec.get('required', [])
-
-    # Protect against the case where a parameter with an inline definition
-    # and a required field is passed to this method.
-    if isinstance(required_fields, bool):
-        required_fields = []
 
     result = {}
     for k, v in iteritems(object_value):
@@ -156,7 +139,6 @@ def marshal_object(swagger_spec, object_spec, object_value):
 
         if v is None and k not in required_fields:
             continue
-
         if prop_spec:
             result[k] = marshal_schema_object(swagger_spec, prop_spec, v)
         else:
@@ -183,11 +165,7 @@ def marshal_model(swagger_spec, model_spec, model_value):
         raise SwaggerMappingError('Unknown model {0}'.format(model_name))
 
     if model_value is None:
-        if is_prop_nullable(swagger_spec, model_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(model_value))
+        return handle_null_value(swagger_spec, model_spec)
 
     if not isinstance(model_value, model_type):
         raise SwaggerMappingError(
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -73,9 +73,12 @@ def marshal_primitive(swagger_spec, primitive_spec, value):
         default_used = True
         value = schema.get_default(swagger_spec, primitive_spec)
 
-    if value is None and schema.is_required(swagger_spec, primitive_spec):
-        raise SwaggerMappingError(
-            'Spec {0} is a required value'.format(primitive_spec))
+    if value is None:
+        if is_prop_nullable(swagger_spec, primitive_spec):
+            return None
+        else:
+            raise SwaggerMappingError(
+                'Spec {0} is a required value'.format(primitive_spec))
 
     if not default_used:
         value = formatter.to_wire(swagger_spec, primitive_spec, value)
@@ -92,6 +95,13 @@ def marshal_array(swagger_spec, array_spec, array_value):
     :rtype: list
     :raises: SwaggerMappingError
     """
+    if array_value is None:
+        if is_prop_nullable(swagger_spec, array_spec):
+            return None
+        else:
+            raise SwaggerMappingError(
+                'Spec {0} is a required value'.format(array_spec))
+
     if not is_list_like(array_value):
         raise SwaggerMappingError('Expected list like type for {0}: {1}'
                                   .format(type(array_value), array_value))
@@ -119,19 +129,32 @@ def marshal_object(swagger_spec, object_spec, object_value):
     """
     deref = swagger_spec.deref
 
+    if object_value is None:
+        if is_prop_nullable(swagger_spec, object_spec):
+            return None
+        else:
+            raise SwaggerMappingError(
+                'Spec {0} is a required value'.format(object_spec))
+
     if not is_dict_like(object_value):
         raise SwaggerMappingError('Expected dict like type for {0}:{1}'.format(
             type(object_value), object_value))
+
+    object_spec = deref(object_spec)
+    required_fields = object_spec.get('required', [])
+
+    # Protect against the case where a parameter with an inline definition
+    # and a required field is passed to this method.
+    if isinstance(required_fields, bool):
+        required_fields = []
 
     result = {}
     for k, v in iteritems(object_value):
 
         prop_spec = get_spec_for_prop(
-            swagger_spec, deref(object_spec), object_value, k)
+            swagger_spec, object_spec, object_value, k)
 
-        # Values cannot be None unless x-nullable is set
-        is_nullable = prop_spec and is_prop_nullable(swagger_spec, prop_spec)
-        if v is None and not is_nullable:
+        if v is None and k not in required_fields:
             continue
 
         if prop_spec:
@@ -158,6 +181,13 @@ def marshal_model(swagger_spec, model_spec, model_value):
 
     if model_type is None:
         raise SwaggerMappingError('Unknown model {0}'.format(model_name))
+
+    if model_value is None:
+        if is_prop_nullable(swagger_spec, model_spec):
+            return None
+        else:
+            raise SwaggerMappingError(
+                'Spec {0} is a required value'.format(model_value))
 
     if not isinstance(model_value, model_type):
         raise SwaggerMappingError(
--- a/bravado_core/unmarshal.py
+++ b/bravado_core/unmarshal.py
@@ -7,7 +7,7 @@ from bravado_core.model import is_model, MODEL_MARKER
 from bravado_core.schema import get_spec_for_prop
 from bravado_core.schema import is_dict_like
 from bravado_core.schema import is_list_like
-from bravado_core.schema import is_prop_nullable
+from bravado_core.schema import handle_null_value
 from bravado_core.schema import SWAGGER_PRIMITIVES
 
 
@@ -72,11 +72,7 @@ def unmarshal_primitive(swagger_spec, primitive_spec, value):
     :raises: SwaggerMappingError
     """
     if value is None:
-        if is_prop_nullable(swagger_spec, primitive_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(primitive_spec))
+        return handle_null_value(swagger_spec, primitive_spec)
 
     value = formatter.to_python(swagger_spec, primitive_spec, value)
     return value
@@ -92,11 +88,7 @@ def unmarshal_array(swagger_spec, array_spec, array_value):
     :raises: SwaggerMappingError
     """
     if array_value is None:
-        if is_prop_nullable(swagger_spec, array_spec):
-            return None
-        else:
-            raise SwaggerMappingError(
-                'Spec {0} is a required value'.format(array_spec))
+        return handle_null_value(swagger_spec, array_spec)
 
     if not is_list_like(array_value):
         raise SwaggerMappingError('Expected list like type for {0}:{1}'.format(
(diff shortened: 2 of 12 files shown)


zroadhouse-rmn:
I'll fix the Travis CI issues with another commit (likely tomorrow or the next day).

zroadhouse-rmn:
There are additional issues in marshal that I am working on as well.

zroadhouse-rmn:
@sjaensch - PR ready for review

sjaensch:
@zroadhouse-rmn this is awesome! Going to review it today, thanks for the contribution!

bravado_core/marshal.py
```diff
@@ -73,9 +73,12 @@ def marshal_primitive(swagger_spec, primitive_spec, value):
         default_used = True
         value = schema.get_default(swagger_spec, primitive_spec)
 
-    if value is None and schema.is_required(swagger_spec, primitive_spec):
-        raise SwaggerMappingError(
-            'Spec {0} is a required value'.format(primitive_spec))
+    if value is None:
+        if is_prop_nullable(swagger_spec, primitive_spec):
+            return None
+        else:
```
sjaensch: Just a minor nit, but I'd remove the else. Raising the exception is a nice default action for this case unless we handle it beforehand. Same on line 101, 136 and 189.

bravado_core/marshal.py
```diff
@@ -73,9 +73,12 @@ def marshal_primitive(swagger_spec, primitive_spec, value):
         default_used = True
         value = schema.get_default(swagger_spec, primitive_spec)
 
-    if value is None and schema.is_required(swagger_spec, primitive_spec):
-        raise SwaggerMappingError(
-            'Spec {0} is a required value'.format(primitive_spec))
+    if value is None:
```
sjaensch: You have this same code block in multiple places. How about putting it into a small helper function? You could do something like

    if value is None:
        return marshal_null_property(swagger_spec, prop_spec, prop_value)

Which then either returns None or raises the exception.

bravado_core/marshal.py
sjaensch: I'd remove that `else`. Raising the exception is a nice final step for this condition, unless we handle it differently beforehand.

bravado_core/marshal.py
```diff
@@ -159,6 +182,13 @@ def marshal_model(swagger_spec, model_spec, model_value):
     if model_type is None:
         raise SwaggerMappingError('Unknown model {0}'.format(model_name))
 
+    if model_value is None:
+        if is_prop_nullable(swagger_spec, model_spec):
+            return None
+        else:
+            raise SwaggerMappingError(
+                'Spec {0} is a required value'.format(model_value))
```
sjaensch: This should be model_spec, not model_value.

bravado_core/unmarshal.py
```diff
@@ -112,14 +120,33 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
...
+    # Protect against the case where a parameter with an inline definition
+    # and a required field is passed to this method.
+    if isinstance(required_fields, bool):
+        required_fields = []
+
     result = {}
     for k, v in iteritems(object_value):
         prop_spec = get_spec_for_prop(
             swagger_spec, object_spec, object_value, k)
+        if v is None and k not in required_fields:
```
sjaensch: Why are you doing this behavior change (i.e. filtering out fields with `None` values)?

tests/marshal/marshal_model_test.py
```diff
@@ -83,3 +84,32 @@ def test_value_is_not_dict_like_raises_error(petstore_dict):
     with pytest.raises(SwaggerMappingError) as excinfo:
         marshal_model(petstore_spec, pet_spec, 'i am not a dict')
     assert 'Expected model of type' in str(excinfo.value)
+
+
+def test_marshal_model_with_none_model_type(petstore_spec):
+    model_spec = {'x-model': 'Foobar'}
+    Pet = petstore_spec.definitions['Pet']
```
sjaensch: Thanks for adding this test! You can just pass in a mock object as pet, it's not required for the test.

bravado_core/marshal.py
```diff
@@ -119,19 +129,32 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
     if not is_dict_like(object_value):
         raise SwaggerMappingError('Expected dict like type for {0}:{1}'.format(
             type(object_value), object_value))
 
+    object_spec = deref(object_spec)
+    required_fields = object_spec.get('required', [])
+
+    # Protect against the case where a parameter with an inline definition
+    # and a required field is passed to this method.
+    if isinstance(required_fields, bool):
```
sjaensch: This shouldn't happen. [Parameter objects](http://swagger.io/specification/#parameterObject) have a required field that has a boolean value; [schema objects](http://swagger.io/specification/#schemaObject) do not, their required field has to be a list. I don't think we should be doing this here, it might be hiding a bug.

tests/marshal/marshal_primitive_test.py
```diff
@@ -39,24 +39,6 @@ def test_skips_default(mock_to_wire, minimal_swagger_spec):
     assert mock_to_wire.call_count == 1
 
 
-def test_required(minimal_swagger_spec):
```
sjaensch: Why are you removing this test?

tests/unmarshal/unmarshal_object_test.py
```diff
@@ -238,42 +239,84 @@ def test_recursive_ref_with_depth_n(recursive_swagger_spec):
...
-    value = {'x': 'y'}
-
+@pytest.mark.parametrize('property_type', ['string', 'object', 'array'])
+def test_nullable_with_value(empty_swagger_spec, nullable, required,
+                             property_type):
+    content_spec = nullable_spec_factory(required, nullable, property_type)
+    if property_type == 'object':
+        value = {'x': {'y': 'z'}}
+    elif property_type == 'array':
+        content_spec['properties']['x']['items'] = {'type': 'string'}
```
sjaensch: Doesn't the nullable_spec_factory already do this?

tests/unmarshal/unmarshal_object_test.py
```diff
@@ -238,42 +239,84 @@ def test_recursive_ref_with_depth_n(recursive_swagger_spec):
...
 @pytest.mark.parametrize('required', [True, False])
-def test_nullable_with_value(empty_swagger_spec, nullable, required):
-    content_spec = nullable_spec_factory(required, nullable)
-    value = {'x': 'y'}
-
+@pytest.mark.parametrize('property_type', ['string', 'object', 'array'])
+def test_nullable_with_value(empty_swagger_spec, nullable, required,
+                             property_type):
+    content_spec = nullable_spec_factory(required, nullable, property_type)
+    if property_type == 'object':
```
sjaensch: I'd prefer having this as part of the parametrization, since that's what it is. Like so:

    @pytest.mark.parametrize(
        'property_type, value',
        ('string', 'y'),
        ('object', {'y': 'z'},
        ('array', ['one', 'two', 'three']),
    )

tests/unmarshal/unmarshal_object_test.py
```diff
@@ -238,42 +239,84 @@ def test_recursive_ref_with_depth_n(recursive_swagger_spec):
...
+    content_spec = nullable_spec_factory(required, True, property_type)
     value = {'x': None}
 
     result = unmarshal_object(empty_swagger_spec, content_spec, value)
     assert result == {'x': None}
+
+
+@pytest.mark.parametrize('property_type', ['string', 'object', 'array'])
+def test_non_nullable_none_value(empty_swagger_spec, property_type):
+    content_spec = nullable_spec_factory(True, False, property_type)
```
sjaensch: It becomes hard to understand what the parameters mean if you pass them in as values. Could you maybe switch to kwargs notation?

bravado_core/marshal.py
zroadhouse-rmn: Will do.

bravado_core/marshal.py
zroadhouse-rmn: Ok.

bravado_core/marshal.py
zroadhouse-rmn: I put that code in there since previously the concepts of parameter-required vs object required were mixed up in the code, and I was unsure if there were any callers that might be passing parameters directly to marshal_object.

I agree that it will obscure issues and will remove it (and the associated test cases).

bravado_core/marshal.py
zroadhouse-rmn: Good catch! Yet another reason to factor that out into a common function.

bravado_core/unmarshal.py
zroadhouse-rmn: I debated on this point (the alternative being to replace continue with result[k] = None). Since the value was None and the property was not required, I skipped setting the property value. The other way makes sense to me as well.

zroadhouse-rmn: Previously I don't believe this behavior was handled properly. This is the only place that the required field can be checked - as unmarshal_schema_object doesn't have sufficient context.

tests/marshal/marshal_model_test.py
zroadhouse-rmn: Will do

tests/marshal/marshal_primitive_test.py
zroadhouse-rmn: I moved them to tests/param/marshal_param_test.py as the required: True/False attribute is only valid for parameters.

tests/unmarshal/unmarshal_model_test.py
```diff
@@ -105,3 +106,131 @@ def test_value_is_not_dict_like_raises_error(petstore_dict):
...
+    assert pet.category is None
+
+
+def test_non_nullable_object_properties(petstore_dict):
+    petstore_dict = copy.deepcopy(petstore_dict)
+    petstore_spec = Spec.from_dict(petstore_dict)
+    pet_spec_dict = petstore_dict['definitions']['Pet']
+    pet_spec_dict['required'].append('category')
+    pet_spec = petstore_spec.spec_dict['definitions']['Pet']
+    pet_dict = {
```
zroadhouse-rmn: I think this is a good suggestion. I copy/pasted for speed, but having a fixture clearly reduces repetition.

tests/unmarshal/unmarshal_object_test.py
zroadhouse-rmn: Didn't know you could do that. I'll try it out.

tests/unmarshal/unmarshal_object_test.py
zroadhouse-rmn: Looks like I missed that when I went back to the tests to update them. I can remove.

tests/unmarshal/unmarshal_object_test.py
zroadhouse-rmn: Totally agree. I debated this point myself.

tests/param/marshal_param_test.py
```diff
@@ -201,3 +201,25 @@ def test_ref(minimal_swagger_dict, array_param_spec, request_dict):
     expected['params']['animals'] = ','.join(value)
     marshal_param(param, value, request_dict)
     assert expected == request_dict
+
+
+def test_required_param(minimal_swagger_spec, string_param_spec, request_dict):
```
zroadhouse-rmn: Here are the tests I moved from tests/marshal/marshal_primitive_test.py

bravado_core/unmarshal.py
zroadhouse-rmn: i.e. Without checking the required fields here, the code would end up in the nullable property check and would fail.

zroadhouse-rmn: Do you think the line deserves a comment?

sjaensch: I think it goes against what we do a couple of lines down, where we explicitly re-introduce missing properties. I think we should keep that behavior. Generally speaking: unmarshalling takes the data transferred over the wire and returns Python dicts (or models). These should have all of the property fields, and if there was no value we use `None`.

zroadhouse-rmn:
@sjaensch - Pushed a new commit based on your code review comments. Thank you for taking the time to carefully review my PR.

bravado_core/unmarshal.py
```diff
@@ -112,15 +112,23 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
...
+    object_spec = deref(object_spec)
+    required_fields = object_spec.get('required', [])
+
     result = {}
     for k, v in iteritems(object_value):
         prop_spec = get_spec_for_prop(
             swagger_spec, object_spec, object_value, k)
-        if prop_spec:
+        if v is None and k not in required_fields:
+            result[k] = None
```
zroadhouse-rmn: Changed this based on our discussion. Makes sense that we should set this explicitly to None here.

bravado_core/unmarshal.py
zroadhouse-rmn: Makes sense. Its changed in the latest commit.

bravado_core/unmarshal.py
sjaensch: Shouldn't we let this be handled by unmarshalling the field? `unmarshal_schema_object()` will call the appropriate function and that will deal with None values - maybe they are not allowed for that property.

tests/marshal/marshal_primitive_test.py
sjaensch: `marshal_primitive()` handles parameters. It uses `schema.is_required()` which checks for `is_required` in that parameter spec, assuming it has a boolean value. Please move those tests back. 
Marshalling converts Python data structures into the wire format for bravado requests, i.e. JSON. These are almost always Parameters, unless it's an object/model (in case of requests with a `body` parameter).

zroadhouse-rmn: Ok. I moved the tests to tests/param/marshal_param_test.py because the code is specific to parameters - the required check is now in marshal_param and not in marshal_primitive. As such, it made more sense to me to have the test cases under the param tests instead of the primitive tests. Still want me to move them back?

tests/param/marshal_param_test.py
zroadhouse-rmn: Note that the tests invoke marshal_param and not marshal_primitive.

sjaensch: Yeah I think it makes sense to have these two unit tests for marshal_primitive. marshal_param will still eventually call marshal_primitive, and since you already wrote them feel free to keep these new tests. But the other two tests were shorter and easier to grasp, which is always a good thing. :)

bravado_core/unmarshal.py
zroadhouse-rmn: No. This condition is here to handle the case when the field is not required and the field is not nullable.
unmarshal_schema_object does not have enough information to make the decision on a required property since required=[] is scoped at the containing type while unmarshal_schema_object has the spec for the property. 

Side note: unmarshal_schema_object can't use the schema.is_required method since that would return the required=[] property for an object type (due to dereferencing) vs. the expected required=bool for a parameter. If the required list is non-null then schema.is_required would evaluate true. Thus we can't use the same process I did for x-nullable properties (i.e. copy of the de-referenced spec and insertion of the x-nullable property) because a boolean required property would overwrite a list property for an object type. Better would be a more extension refactoring (which I didn't have time for) where the marshaling code is also passed the type stack and can walk up the chain to check properties like required=[] and x-nullable.

I was missing a test case for this particular combination. I'll add one.

bravado_core/marshal.py
zroadhouse-rmn: Done - this is now called handle_null_value since it is used in both marshal and unmarshal pathways.

tests/param/marshal_param_test.py
zroadhouse-rmn: The tests will fail now if moved back to marshal_primitive since the schema.is_required check has been moved to marshal_param.
I apologize if the newer tests are harder to read. I tried to keep them consistent with the other parameter tests.

bravado_core/unmarshal.py
zroadhouse-rmn: Added tests/unmarshal/unmarshal_object_test.py::test_non_required_none_value to cover the case.

sjaensch: Ok, I see now why this is pretty much a mess: I agree with you that the original code in `unmarshal_primitive()` was problematic: the use of `schema.is_required` did not work if the passed spec was not for a parameter.

The core issue here is that unmarshalling is used both for the request data (server side) as well as the response from the server (client side). Both use `unmarshal_schema_object()`, for request unmarshalling it's called by `unmarshal_param()`. There is a helper function `is_param_spec()` to distinguish between the two.

We can still access the required field even for model properties: this is why we pass in the swagger spec in addition to the object spec when unmarshalling.

It looks like `required` is only correctly handled during validation on the client side, which is all that was really needed previously. I'd say we stick to the minimal change we need for this to work, in order not to make this PR even bigger.

zroadhouse-rmn: Agree that I don't want to make this PR bigger. Is there anything else you'd like me to revise to have it be accepted? I think its in a good state now.

sjaensch: Absolutely @zroadhouse-rmn , thanks for being so patient and improving on this CR! Let's merge it in and get some real-world testing. 😁

zroadhouse-rmn: 👍 Thank you!

sjaensch: You're welcome @zroadhouse-rmn, I'll make sure to call you out in the release notes!
I've just discovered another issue and posted #111. Since you've spent quite a bit of time on that area of the code, would you mind giving it a quick review and let me know what you think?

PR #111, opened 2016-08-16
Fix AttributeError if required array param is not supplied

sjaensch:
We've observed AttributeErrors internally if a required parameter of type array is not passed, instead of the expected schema or validation exception. This is because validation happens after unmarshalling of the array, and a missing required parameter is not currently handled correctly during unmarshalling.


zroadhouse-rmn:
Looks good! Nice to see the handle_null_value get put to immediate use.

PR #112, opened 2016-08-23
Security Objects handling and parameter validation

macisamuele:
Handling of security definitions for an Operation object and enforcing request parameter validation  in case of ``apiKey`` security.


bravado_core/operation.py
```diff
@@ -111,6 +147,28 @@ def __repr__(self):
         return u"%s(%s)" % (self.__class__.__name__, self.operation_id)
 
 
+def _build_params_from_security_objects(op):
+    """
+    Generate the required parameters from the security object definition.
+    NOTE: the current implementation handles only basic and apiKey types.
```
laucia: we should probably start mentioning this in the documentation. 
Could you add an entry for security support in there?

macisamuele: sorry @laucia but is not clear to me what you meant ... could you please give me more informations? Thanks a lot

bravado_core/operation.py
```diff
@@ -94,7 +129,8 @@ def sanitize(operation_id):
             # Handle crazy corner cases where someone explictily sets operation
             # id a value that gets sanitized down to an empty string
             if len(operation_id) == 0:
-                operation_id = sanitize(self.http_method + '_' + self.path_name)
+                operation_id = sanitize(
+                    self.http_method + '_' + self.path_name)
```
sjaensch: Why was this necessary? There shouldn't be a 80 character line length limit anymore.

tests/request/unmarshal_request_test.py
```diff
@@ -6,7 +6,11 @@
...
 def test_request_with_path_parameter(petstore_spec):
-    request = Mock(spec=IncomingRequest, path={'petId': '1234'})
+    request = Mock(
+        spec=IncomingRequest,
+        path={'petId': '1234'},
+        headers={'api_key': 'key1'},
+    )
     # /pet/{pet_id} fits the bill
     op = petstore_spec.resources['pet'].operations['getPetById']
     request_data = unmarshal_request(request, op)
```
sjaensch: Could you add an assert here that makes sure the api_key in in the request data as well?

sjaensch:
This is looking really good! Could you please add a test or two making sure the new security object parameters behave properly for validation? I.e. we'd want a validation error if they're missing for example, at least for the api_key type.

bravado_core/operation.py
macisamuele: The constraint was enforced by the configurations on ``.pre-commit-config.yaml``.
I have updated the file in order to allow rows up to 131 characters.

macisamuele:
I have added 2 tests ``test_correct_request_with_apiKey_security`` and ``test_wrong_request_with_apiKey_security`` for checking the proper validation behaviour.

tests/request/unmarshal_request_test.py
macisamuele: Added the assertion

sjaensch:
👍 Taking this, it's a nice foundation to build upon.

macisamuele:
@sjaensch Thanks a lot. In the next days I'll open a new PR for the documentation as suggested by @laucia.

PR #113, opened 2016-08-31
Give Models an easier way to roundtrip

thebostik:
Provide access to the existing marshal and unmarshal implementations on the Model objects to help us serialize / deserialize more easily without needing to keep track of specs. 

ref: COREBACK-452 Bravado model object responses can't be cached with memcached because it cannot be serialized by pickle


sjaensch:
Merged in #120, which is this change plus a test and a fix.

Issue #116, opened 2016-09-09
build_resource failure for root path with no tags


macisamuele:
The issue is related to what is reported in [bravado issue #239](https://github.com/Yelp/bravado/issues/239).

An small example of the issue is
```python
Python 2.7.9 (default, Aug 23 2016, 10:24:57)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from bravado_core.spec import Spec
>>> import yaml
>>>
>>> spec_text = """
info:
...   version: 0.0.0
...   title: Simple API
... paths:
...   /:
...     get:
...       responses:
...         '200':
...           description: OK
... swagger: '2.0'
... """
>>> spec_dict = yaml.load(spec_text)
>>> print spec_dict
{'info': {'version': '0.0.0', 'title': 'Simple API'}, 'paths': {'/': {'get': {'responses': {'200': {'description': 'OK'}}}}}, 'swagger': '2.0'}
>>> swagger_spec = Spec.from_dict(spec_dict)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/maci/pg/bravado/.tox/py27/lib/python2.7/site-packages/bravado_core/spec.py", line 152, in from_dict
    spec.build()
  File "/Users/maci/pg/bravado/.tox/py27/lib/python2.7/site-packages/bravado_core/spec.py", line 175, in build
    self.resources = build_resources(self)
  File "/Users/maci/pg/bravado/.tox/py27/lib/python2.7/site-packages/bravado_core/resource.py", line 67, in build_resources
    tags.append(convert_path_to_resource(path_name))
  File "/Users/maci/pg/bravado/.tox/py27/lib/python2.7/site-packages/bravado_core/resource.py", line 31, in convert_path_to_resource
    raise SwaggerMappingError(err_msg.format(path_name))
bravado_core.exception.SwaggerMappingError: Could not extract resource name from path /
```

The issue seems caused by the assumption that the path will match the following regexp ``/.+`` ([resource.py](https://github.com/Yelp/bravado-core/blob/2bcb9e1d452931e683b0c3ff7c30a9bd2ae06fa2/bravado_core/resource.py#L29))

jkrukoff:
Just ran into this issue. Was fortunate enough to also be in control of the swagger spec, so as to be able to add a superfluous tag to my root path.

PR #117, opened 2016-09-09
Handle root path with no tag resource generation

macisamuele:
The PR would relax a little the constraint of the resource building process (issue #116).
At the moment the resource building requires either the path has a `tag` or the path matches the following regex `/.+`.

I believe that the constraint could be relaxed in order to allow the correct validation of specs like this

``` json
{
    "info": {
        "version": "0.0.0",
        "title": "Simple API"
    },
    "paths": {
        "/": {
            "get": {
                "responses": {
                    "200": {
                        "description": "OK"
                    }
                }
            }
        }
    },
    "swagger": "2.0"
}
```

### DIFF
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -3,7 +3,6 @@ import logging
 
 from six import iteritems
 
-from bravado_core.exception import SwaggerMappingError
 from bravado_core.operation import Operation
 
 log = logging.getLogger(__name__)
@@ -15,6 +14,7 @@ def convert_path_to_resource(path_name):
     name on a best effort basis when an operation has no tags.
 
     Examples:
+      /                   ->  _
       /pet                ->  pet
       /pet/findByStatus   ->  pet
       /pet/findByTags     ->  pet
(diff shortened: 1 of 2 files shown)


sjaensch:
Hmm, we have a test that made sure we throw an error for the '/' case. Are you absolutely positive this is allowed by the spec? Also, why did you remove the test for the '' case?

macisamuele:
Basically the resource attribution is provided by the defined tags (which are optional) or by the "first part" of the url. I think that this assumption could be too restrictive because a route like: ``/`` or ``/{id}`` MUST have defined a tag to be used and clients could not have permissions to modify specs.

To relax a little this situation I would allow a default tag ``_`` for that cases.
In case the issue is related to the choose of ``_`` (the "default" throwaway python variable) we could think to move it to ``default`` which is the solution adopted by [swagger-ui](http://petstore.swagger.io) (with swagger specs from [http://editor.swagger.io/spec-files/minimal.yaml](http://editor.swagger.io/spec-files/minimal.yaml))

PR #120, opened 2016-09-27
Add marshal and unmarshal to bravado-core.model's and testing

bschmaltz:
Add marshal and unmarshal functionality, so we can deserialize and serialize easily in frontend services.

Same as this PR - https://github.com/Yelp/bravado-core/pull/113  But, I added tests, and found the unmarshal lambda function needed to be wrapped with staticmethod for python 2.6 and 2.7.  Otherwise, I'd get "TypeError: unbound method <lambda>()..."

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -82,6 +82,8 @@ def create_model_type(swagger_spec, model_name, model_spec):
     :returns: dynamic type created with attributes, docstrings attached
     :rtype: type
     """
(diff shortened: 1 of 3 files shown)


macisamuele:
lg2m
Thanks @bschmaltz and @thebostik for the idea and the contribution.

sjaensch:
Awesome, thanks @bschmaltz! I'm glad that writing those tests helped us prevent bugs. Going to merge this in.

PR #121, opened 2016-09-27
Allow overriding security spec with empty list

jlumpe:
Currently the global `security` property of the spec can be overridden in specific operations, but only if the list of security requirements is non-empty.

In the [official spec](http://swagger.io/specification/#operationObject) it states that

>  To remove a top-level security declaration, an empty array can be used

for an operation's `security` property.

This change should bring things in line with the official spec.

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -59,8 +59,9 @@ class Operation(object):
             deref = self.swagger_spec.deref
             op_spec = deref(self.op_spec)
             spec_dict = deref(self.swagger_spec.spec_dict)
-            security_spec = deref(op_spec.get('security', []))
-            if len(security_spec) == 0:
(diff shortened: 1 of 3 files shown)


macisamuele:
@jlumpe thanks a lot for your contribution and for having identified the specs un-compliant behaviour.

Your change looks correct to me but I would ask if you can add a couple of tests in [security_object_test.py](https://github.com/Yelp/bravado-core/blob/master/tests/operation/security_object_test.py) that validates your change.

jlumpe:
Updated with new test. Test passes this branch, fails on master. Hope I got the code style right.

macisamuele:
@jlumpe: thanks for the tests ... with them we could be sure that the code change is doing what is expected.

PR #122, opened 2016-09-28
Contribution guidelines and master changelog

macisamuele:
As done in [bravado](https://github.com/Yelp/bravado) could be helpful having a sort of guideline for the project contribution.

The CHANGELOG-MASTER.rst file is added in order to avoid to have the changelog related to the master modifications on readthedocs.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,10 @@
 Changelog
 =========
+X.X.X (XXXX-XX-XX)
+------------------
+- The content will be used to build the Changelog for the new bravado-core release
+  (add before this line your modifications summary and PR reference)
+
 4.5.1 (2016-09-27)
 ------------------
 - Add marshal and unmarshal methods to models - PR #113, #120.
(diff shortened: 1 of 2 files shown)


sjaensch:
Upon further consideration, I'm not sure splitting the changelog into two files is such a good idea. The documentation for the newest stable version shouldn't be affected by this - is this working correctly, i.e. is readthedocs finding our tags? I think so.

With this change the changelog on readthedocs would not be up to date anymore even though the version says "latest" (i.e. current master). Thoughts?

macisamuele:
As far as I know the latest version available by readthedocs is in any case a published version and not the one on master.
For me having the changelog-master file is:
 * a way to keep distinction between what is published and what is currently on master
 * a workaround to avoid having a "fake" version information in the readthedocs page

Is it correct? In any case we can re-think how to simplify the contribution process :D

macisamuele:
@sjaensch I have deleted ``CHANGELOG-MASTER.rst`` file. The content is included in the main changelog file since it will be shown in readthedocs for the latest version (which is pointing to our master). I have also updated the ``README.rst``.

sjaensch:
👍 lg2m

Issue #123, opened 2016-09-28
Fix handling of security requirements to match spec


jlumpe:
For both the root swagger object and an operation object the `security` property is defined to be an array of [security requirement objects](http://swagger.io/specification/#securityRequirementObject). The official spec says the following in the description for this property:

> The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements).

The security requirement object is described as:

> Lists the required security schemes to execute this operation. The object can have multiple security schemes declared in it which are all required (that is, there is a logical AND between the schemes).

The current implementation effectively just uses a logical AND between all security schemes in all security requirements by marking all parameters derived from these as required. This prevents alternate sets of security schemes from being used. It also introduces a possible bug where parameters would be duplicated if the same security scheme appears in two or more different security requirement objects.

I can submit a pull request which removes the `required` property from security parameters and removes possible duplicates from the `Operation.security_objects` property. However this is not a complete fix as no checks are performed to make sure a valid set of security parameters are given in a request. I'm not sure that fix belongs in this package, probably it should be done in `bravado.client.construct_params` instead.

macisamuele:
@jlumpe Thanks a lot for pointing out this!
In the next days I'll work on it. It is important to be as strict as possible to the specs 😄 

I would ask you to confirm if my understanding is correct.

According to the swagger specs below:
* the parameters defined for ``/example1`` are ``sec1`` as required
* the parameters defined for ``/example2`` are ``sec1`` as required and ``sec2`` as optional. Well defined requests have ``sec1`` only or ``sec1`` and ``sec2`` in the header.
* the parameters defined for ``/example2`` are ``sec1`` as required and ``sec2``, ``sec3`` as optional. Well defined requests have ``sec1`` only or ``sec1`` and ``sec2`` or ``sec1`` and ``sec3``  in the header. ``sec2`` and ``sec3`` cannot be present at the same time in the header.
```json
{
  "swagger": "2.0",
  "info": {"title": "Test","version": "1.0"},
  "securityDefinitions": {
    "sec1": {
      "type": "apiKey",
      "name": "sec1",
      "in": "header"
    },
    "sec2": {
      "type": "apiKey",
      "name": "sec2",
      "in": "header"
    },
    "sec3": {
      "type": "apiKey",
      "name": "sec3",
      "in": "header"
    }
  },
  "security": [
    {
      "sec1": []
    }
  ],
  "paths": {
    "/example1": {
      "get": {
[...]

PR #124, opened 2016-10-06
Make Security Definition and Security Requirement as Python Objects

macisamuele:
The goal of this PR is to target #123.

As @jlumpe the initial implementation of the security object added by PR #112 was not 100% compliant with the swagger specs. 

To achieve a better matching of the specs I have created specific classes for [Security Definition](http://swagger.io/specification/#securityDefinitionsObject) and [Security Requirement](http://swagger.io/specification/#securityRequirementObject). In this way will be easier to integrate the handling of ``oauth2`` security format.

Main differences:
* ``security_objects`` property of ``operation`` is removed: the definition was incomplete and is replaced by ``security_requirements`` property which returns the list of Security Requirement objects
* the parameters injected due to security specification cannot override parameters required by Path Object or Operation Object (is needed because different constraint could be injected) and are marked as optional parameters (because of the OR between different security requirements, check ``example3`` on issue #123)
* ``unmarshal_request`` performs proper validation of security related parameters:
    * checks that 1 and only one security requirement is matched
    * in case of multiple security requirements are matched, checks that one matching security requirement is a superset of all the others (contains all the others, ie. ``example5`` on issue #123)

I know that reviewing it could be hard ... so I added a _long_ set of tests to guarantee a good coverage of the modifications.

**NOTE**: I'll update ``CHANGELOG.rst`` after the end of the review


jlumpe:
Can confirm that this is working for me using my simple spec where the API key can be in the query or header.

macisamuele:
I have added test care with apiKey security with parameter in query.

@jlumpe the test modified is checking even that case.

macisamuele:
@sjaensch if you have some time in the next days could you have a look to this? Thanks a lot

sjaensch:
🛳

Issue #125, opened 2016-10-10
Creation of models with property named "self" fails.


jkrukoff:
When attempting to invoke an endpoint that creates a model with a "self" property, the following traceback is emitted:

```
Traceback (most recent call last):
  File "<ipython-input-14-acee5e06b3f1>", line 1, in <module>
    try: c.index.get_index().result()._links.self.href
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado/http_future.py", line 77, in result
    self.response_callbacks)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado/http_future.py", line 115, in unmarshal_response
    operation)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/response.py", line 110, in unmarshal_response
    op.swagger_spec, content_spec, content_value)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 51, in unmarshal_schema_object
    return unmarshal_model(swagger_spec, schema_object_spec, value)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 173, in unmarshal_model
    model_as_dict = unmarshal_object(swagger_spec, model_spec, model_value)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 133, in unmarshal_object
    result[k] = unmarshal_schema_object(swagger_spec, prop_spec, v)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 51, in unmarshal_schema_object
    return unmarshal_model(swagger_spec, schema_object_spec, value)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 173, in unmarshal_model
    model_as_dict = unmarshal_object(swagger_spec, model_spec, model_value)
  File "/home/jkrukoff/Documents/Services/unifi-go/.virtualenv/lib/python3.5/site-packages/bravado_core/unmarshal.py", line 133, in unmarshal_object
[...]

PR #126, opened 2016-10-10
[allow-model-self-property]: Allow models to have a "self" property.

jkrukoff:
This adds support for models to contain properties named "self", without
triggering a name clash in the `__init__` method.

Unmarshaling is changed to use a new constructor in order to avoid
changing the signature of the `__init__` method.

I'm unsure how to simplify the test case, as how bravado chooses to create models is unclear to me. Worked towards my use case until was able to replicate a model with a self property.

In reference to issue #125

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,7 @@ Changelog
 =========
 X.X.X (XXXX-XX-XX)
 ------------------
+- Allow self as name for model property, adds new "create" alternate model constructor - Issue #125, PR #126.
 - Allow overriding of security specs - PR #121
 - The content will be used to build the Changelog for the new bravado-core release
   (add before this line your modifications summary and PR reference)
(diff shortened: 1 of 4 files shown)


jkrukoff:
It may be possible to do something as simple as change the name of the "self" parameter in __init__, but I don't know of any value there that wouldn't feel like just kicking the can down the road for somebody else to run into. Really only something truly ugly like `__self_{uuid}` seems likely to avoid future conflicts.

macisamuele:
@jkrukoff thanks a lot for your contribution. Definitely that case needed to be handled.
I would ask you if you can add a summary of your changes in the ``CONTRIBUTIONS.rst`` file as shown in our [contributing](https://github.com/Yelp/bravado-core/#contributing) section. It will allow to have a clean history 😄

jkrukoff:
Added summary to CHANGELOG.rst

PR #128, opened 2016-10-14
Have models inherit from base class, support model polymorphism

jlumpe:
Re-worked model generation to use a base `Model` class that defines most of the functionality. The current implementation subclasses `object` and adds lambdas with closures to the `__dict__` in lieu of methods. This allows for custom extensions as well as I think being cleaner and more Pythonic.

One issue with dynamic generation of model classes is that the schema properties could be named anything so there will always be possible conflicts with the base model methods/attributes. I've attempted to minimize this by prefixing all class-level attribute names that are part of the public API with a single underscore. This is unusual but is similar to the `collections.namedtuple` class which also creates dynamic classes with arbitrary attribute names. Truly private attributes are prefixed with a double underscore.

Quick list of changes:
- Property values stored in `Model.__dict` attribute instead of standard `Model.__dict__` so that attribute access can be customized.
- Attribute access prioritizes class attributes (methods, etc.) over model properties, so that the public API will never be accidentally hidden. This is only an issue with property names beginning with underscores.
- Property values can be accessed explicitly using `model['prop_name']`. This works for setting and deleting too.
- Deleting a property defined in the spec (through `del model.property` or `del model['property']`) will set it to `None` instead of deleting it entirely.
- Models function as a container of property names for `__iter__` and `__contains__` (similar to a dictionary).
- `Model._additional_props` is now dynamic.
- Docstrings are still generated lazily.
- `marshal()` and `unmarshal()` methods renamed `_marshal()` and `_unmarshal()`.
- `_to_dict()` gets dictionary of property values.
- `create_model_type()` allows using a custom base class.
- Sorry but I think I overwrote the newest fix allowing "self" properties. The `Model._from_dict()` method
  is pretty much the same thing as the new `create()`.

Also, polymorphism is now implemented as defined as per the Swagger 2.0 spec. If a model spec has the `discriminator` property defined, this value will be checked when unmarshaling data to choose an alternative model.

TODO:
- Write tests for new functionality
- Instantiating with additional properties when `schema['additionalProperties'] == False` raises an exception, but no checking is done when setting attributes afterward.
- I think `Model.__dir__()` should be more consistent with the standard implementation and return private and class attributes as well. To get the current property names you can just iterate over it. For now I've kept like it was before to pass the existing test.
- `Model.__repr__()` can get very long, maybe truncate it?
- Validate properties when they are set?

### DIFF
--- a/bravado_core/docstring.py
+++ b/bravado_core/docstring.py
@@ -8,15 +8,6 @@ from bravado_core.exception import SwaggerMappingError
 log = logging.getLogger(__name__)
 
 
-class docstring_property(object):
-
-    def __init__(self, func):
-        self.func = func
-
-    def __get__(self, _cls, _owner):
-        return self.func()
-
-
 def operation_docstring_wrapper(operation):
     """
     Workaround for docstrings to work for operations in a sane way.
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -167,7 +167,7 @@ def marshal_model(swagger_spec, model_spec, model_value):
     if model_value is None:
         return handle_null_value(swagger_spec, model_spec)
 
-    if not isinstance(model_value, model_type):
+    if not model_type._isinstance(model_value):
         raise SwaggerMappingError(
             'Expected model of type {0} for {1}:{2}'
             .format(model_name, type(model_value), model_value))
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -174,9 +174,6 @@ def marshal_model(swagger_spec, model_spec, model_value):
 
     # just convert the model to a dict and feed into `marshal_object` because
     # models are essentially 'type':'object' when marshaled
-    attr_names = dir(model_value)
-    object_value = dict(
-        (attr_name, getattr(model_value, attr_name))
-        for attr_name in attr_names)
+    object_value = model_value._as_dict()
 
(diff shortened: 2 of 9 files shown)


jlumpe:
I can write some additional tests for this, but it would be good to get feedback on what I've got first.

sjaensch:
@jlumpe this is looking really good! Could you add a couple of tests for model polymorphism?

Also, do you consider all of your TODOs blockers for shipping this? Some of those seem like separate issues, if you agree please file them accordingly.

jlumpe:
@sjaensch Thanks for reviewing this. I would say other than the tests everything in the TODO list is more of a "nice to have" and can be done in separate PRs later.

If this all looks good I can get started on writing the remaining tests right away. Just wanted to confirm that the API is acceptable, I know prefixing non-private methods with an underscore is unconventional but I don't see much of a way around it here.

Also, it looks like all the Travis checks for Python 2.6 keep failing due to the use of some 2.7-specific syntax. I'll have to clean that up as well.

sjaensch:
@jlumpe honestly I don't have a better solution that makes sure we won't have name clashes with model fields, so let's do it.

macisamuele:
@jlumpe I took your changes and added tests for discriminator feature in #159.
If you feel that something else is missing please let us to know.
And thanks a lot for your great contribution on this project ... it is really appreciated!

PR #130, opened 2016-10-27
129 - additional properties is false by default; raises exception if …

dedominicisfa:
# …unspecified property is marshalled


macisamuele:
Closing this as really old and with little to none context provided on why this change should be done

PR #135, opened 2016-12-13
Some changes which enabled me to use the PagerDuty swagger

jakedt:


advance512:
This is a great contribution, but unfortunately the build fails and tests are missing, decreasing coverage.

@jakedt, would you mind adding proper tests so this can be merged in? It is highly useful.

This is referenced in issue #7.

sjaensch:
@jakedt thanks for your contribution! Would you be able to add a few tests that verify the new behavior? Ignore the py26 test failure for now, I think we'll remove Python 2.6 support real soon now.

PR #138, opened 2016-12-17
Build origin file URIs correctly in Windows

jlumpe:
Several tests were failing in Windows because they built file URIs like so:

    from six.moves.urllib import parse as urlparse
    origin_url = urlparse.urljoin('file:', abs_path_to_spec)

This works for POSIX-style paths but produces invalid URIs for Windows, leading to `jsonschema.validators.RefResolver` being unable to find references relative to a file path. The solution is to use the `urllib.request.pathname2url` method on the absolute path before joining it with the `file://` scheme.

Since using a file for the `origin_url` argument to `Spec` is very common, I added an alternative `origin_file` argument that constructs the correct `origin_url` for you.

Note that I have changed the signature of `Spec.from_dict`, it now takes a single `spec_dict` argument and all keyword arguments are simply passed along to `Spec.__init__`.  This makes it less fragile but could be a breaking change for anyone relying on `origin_url` being the second positional argument, for example.


bravado_core/spec.py
```diff
@@ -55,24 +56,52 @@
...
+                 http_client=None, config=None):
         self.spec_dict = spec_dict
-        self.origin_url = origin_url
         self.http_client = http_client
         self.api_url = None
         self.config = dict(CONFIG_DEFAULTS, **(config or {}))
 
+        if origin_url is not None:
+            if origin_file is not None:
+                raise TypeError(
```
laucia: Could you please add a test for this error?

jlumpe: Done!

jlumpe:
Added one additional test for passing both `origin_url` and `origin_file` to `Spec.from_dict`. Also allowed `origin_file` to be relative and just call `os.path.abspath` on it.

kamalgill:
Ran into this issue testing a swagger spec on Windows using pyramid_swagger.  Any chance we could resolve the merge conflicts to move this along?

Issue #133, opened 2016-12-01
date-time to_wire doesn't validate against rfc3339


asottile:
https://github.com/Yelp/bravado-core/blob/24461141a308eeca4f997b48b4c4deacaef8c12a/bravado_core/formatter.py#L114

```python
>>> strict_rfc3339.validate_rfc3339(datetime.datetime.utcnow().isoformat())
False
```

The "simplest" fix is to do `.isoformat() + 'Z'`, but either way it's impossible to have correct timezone information with just a `datetime` value

asottile:
If dependencies are correctly honored (through `jsonschema[format]`) it makes it impossible to round trip these values

macisamuele:
A valid RFC 3339 format needs to have the timezone information. The timezone could be provided by hour offset (``-08:00`` in case of PST time) or by a letter (such as ``Z`` for ``UTC``).

To have the correct behaviour is important to have a localized datetime object (ie. ``pytz.timezone('Los_Angeles').localize(<näive datetime object>)``).
```python
maci $ ipython
Python 2.7.12 (default, Oct 11 2016, 05:20:59)
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import strict_rfc3339

In [2]: import datetime

In [3]: import pytz

In [4]: strict_rfc3339.validate_rfc3339(datetime.datetime.now().isoformat())
Out[4]: False

In [5]: strict_rfc3339.validate_rfc3339(pytz.utc.localize(datetime.datetime.now()).isoformat())
Out[5]: True
```

PR #140, opened 2016-12-19
date-time formatter to-wire defaults to UTC timezone

macisamuele:
As highlighted on #133 ``date-time`` to-wire function generates invalid result in case of näive datetime objects.
To workaround this issue I'm defaulting the datetime timezone to UTC (in case it is not present).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,3 +2,4 @@
 mock<1.1.0
 pre-commit
 pytest
+pytz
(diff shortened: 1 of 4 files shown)


macisamuele:
@laucia thanks a lot for the fast review!

requirements-dev.txt
```diff
@@ -2,3 +2,4 @@
 mock<1.1.0
 pre-commit
 pytest
+pytz
```
laucia: this is in setup.py too

macisamuele: @asottile++ I simply forgot to remove it! thanks for the fix

PR #141, opened 2016-12-19
Remove prod deps from requirements-dev.txt

asottile:
"Unit test dependencies" at the top of the file


laucia:
Arguably the unittests are using pytz directly, maybe we should clarify the comment to Extra Unit Test dependencies ?

asottile:
Personally, I'd remove the comment.  It's a very common convention and should be obvious from the filename.

PR #144, opened 2017-02-06
[RFC] Test processing of binary body of request

lucafavatella:
I am trying to use bravado for generating a python client from a swagger 2.0 specification that has a PUT request with body whose `Content-Type` header is arbitrary and body is arbitrary sequence of bytes (type string, format binary). The client gets generated but when issuing request the client throws exception. I distilled exception to failing tests in bravado-core - this PR - and I welcome comments. The impression I have is that on Python 3 bravado forces body to be valid Unicode sequence, but I am not sure.

The 3 test failures are similar - the first one is:
```
=================================== FAILURES ===================================
________________________ test_body_binary_octet_stream _________________________

empty_swagger_spec = <bravado_core.spec.Spec object at 0x10e739a90>
param_spec = {'description': 'ID of pet that needs to be fetched', 'in': 'body', 'name': 'petId', 'schema': {'format': 'binary', 'type': 'string'}}

    def test_body_binary_octet_stream(empty_swagger_spec, param_spec):
        param_spec['in'] = 'body'
        param_spec['schema'] = {
            'type': 'string',
            'format': 'binary'
        }
        del param_spec['type']
        del param_spec['format']
        request = {
            'headers': {
                'Content-Type': 'application/octet-stream'
            }
        }
>       marshal_param(param, bytes(10), request)

tests/param/marshal_param_test.py:174:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
bravado_core/param.py:122: in marshal_param
    validate_schema_object(swagger_spec, param_spec, value)
bravado_core/validate.py:24: in validate_schema_object
    validate_primitive(swagger_spec, schema_object_spec, value)
bravado_core/validate.py:49: in validate_primitive
    resolver=swagger_spec.resolver).validate(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <jsonschema.validators.create.<locals>.Validator object at 0x10e46fe10>
args = (b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00',), kwargs = {}
error = <ValidationError: "b'\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00' is not of type 'string'">

    def validate(self, *args, **kwargs):
        for error in self.iter_errors(*args, **kwargs):
>           raise error
E           jsonschema.exceptions.ValidationError: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' is not of type 'string'
E
E           Failed validating 'type' in schema:
E               {'format': 'binary', 'type': 'string'}
E
E           On instance:
E               b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'

.tox/py36/lib/python3.6/site-packages/jsonschema/validators.py:130: ValidationError
```

----
Original exception using bravado:
```
**********************************************************************
Exception raised:
      File "/Users/luca/homebrew/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/python3.6/doctest.py", line 1330, in __run
        compileflags, 1), test.globs)
...
      File "<FILE.rst[13]>", line 2, in FUNCTION
        return CLIENT.TAG.OP(..., body=b, _request_options={'Content-Type': ct
      File "/Users/luca/dev/.../src/bravado/bravado/client.py", line 250, in __call__
        self.operation, request_options, **op_kwargs)
      File "/Users/luca/dev/.../src/bravado/bravado/client.py", line 291, in construct_request
        construct_params(operation, request, op_kwargs)
      File "/Users/luca/dev/.../src/bravado/bravado/client.py", line 313, in construct_params
        marshal_param(param, param_value, request)
      File "/Users/luca/homebrew/lib/python3.6/site-packages/bravado_core/param.py", line 122, in marshal_param
        validate_schema_object(swagger_spec, param_spec, value)
      File "/Users/luca/homebrew/lib/python3.6/site-packages/bravado_core/validate.py", line 24, in validate_schema_object
        validate_primitive(swagger_spec, schema_object_spec, value)
      File "/Users/luca/homebrew/lib/python3.6/site-packages/bravado_core/validate.py", line 49, in validate_primitive
        resolver=swagger_spec.resolver).validate(value)
      File "/Users/luca/homebrew/lib/python3.6/site-packages/jsonschema/validators.py", line 123, in validate
        raise error
    jsonschema.exceptions.ValidationError: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' is not of type 'string'

    Failed validating 'type' in schema:
        {'format': 'binary', 'type': 'string'}

    On instance:
        b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

### DIFF
--- a/bravado_core/swagger20_validator.py
+++ b/bravado_core/swagger20_validator.py
@@ -7,7 +7,11 @@ from jsonschema.validators import Draft4Validator
 from bravado_core.schema import is_param_spec
 from bravado_core.schema import is_prop_nullable
 from bravado_core.schema import is_required
+from bravado_core.schema import has_format
+from bravado_core.schema import get_format
 from swagger_spec_validator.ref_validators import in_scope
+
+import six
 
 """Draft4Validator is not completely compatible with Swagger 2.0 schema
 objects like parameter, etc. Swagger20Validator is an extension of
@@ -39,8 +43,14 @@ def type_validator(swagger_spec, validator, types, instance, schema):
             is_prop_nullable(swagger_spec, schema)) and instance is None:
         return
 
-    for error in _validators.type_draft4(validator, types, instance, schema):
-        yield error
+    if has_format(swagger_spec, schema) and (get_format(swagger_spec, schema) == 'binary') and six.PY3:
+        if not isinstance(instance, bytes):
+            yield ValidationError("bad type")
+        else:
+            return
+    else:
+        for error in _validators.type_draft4(validator, types, instance, schema):
+            yield error
 
 
 def required_validator(swagger_spec, validator, required, instance, schema):
--- a/tests/param/marshal_param_test.py
+++ b/tests/param/marshal_param_test.py
@@ -157,6 +157,63 @@ def test_body(empty_swagger_spec, param_spec):
     assert APP_JSON == request['headers']['Content-Type']
 
 
+def test_body_binary_octet_stream(empty_swagger_spec, param_spec):
+    param_spec['in'] = 'body'
+    param_spec['schema'] = {
+        'type': 'string',
+        'format': 'binary'
+    }
+    del param_spec['type']
+    del param_spec['format']
+    param = Param(empty_swagger_spec, Mock(spec=Operation), param_spec)
+    request = {
+        'headers': {
+            'Content-Type': 'application/octet-stream'
+        }
+    }
+    marshal_param(param, bytes(10), request)
+    assert b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' == request['data']
+    assert 'application/octet-stream' == request['headers']['Content-Type']
+
+
+def test_body_binary_json(empty_swagger_spec, param_spec):
+    param_spec['in'] = 'body'
+    param_spec['schema'] = {
+        'type': 'string',
+        'format': 'binary'
+    }
+    del param_spec['type']
+    del param_spec['format']
+    param = Param(empty_swagger_spec, Mock(spec=Operation), param_spec)
+    request = {
+        'headers': {
+            'Content-Type': 'application/json'
+        }
+    }
+    marshal_param(param, bytes(10), request)
+    assert b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' == request['data']
+    assert 'application/json' == request['headers']['Content-Type']
+
+
+def test_body_binary_foo(empty_swagger_spec, param_spec):
+    param_spec['in'] = 'body'
+    param_spec['schema'] = {
+        'type': 'string',
+        'format': 'binary'
+    }
+    del param_spec['type']
+    del param_spec['format']
+    param = Param(empty_swagger_spec, Mock(spec=Operation), param_spec)
+    request = {
+        'headers': {
+            'Content-Type': 'application/foo'
+        }
+    }
+    marshal_param(param, bytes(10), request)
+    assert b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' == request['data']
+    assert 'application/foo' == request['headers']['Content-Type']
+
+
 def test_formData_integer(empty_swagger_spec, param_spec):
     param_spec['in'] = 'formData'
     param = Param(empty_swagger_spec, Mock(spec=Operation), param_spec)


lucafavatella:
I think I need to somehow modify [here](https://github.com/Yelp/bravado-core/blob/b192c199c7e7303a03ce0e07e11c5bfec25e9d2b/bravado_core/swagger20_validator.py#L152).

lucafavatella:
With the draft patch I wrote the error changed to:
```
________________________ test_body_binary_octet_stream _________________________

empty_swagger_spec = <bravado_core.spec.Spec object at 0x1107759b0>
param_spec = {'description': 'ID of pet that needs to be fetched', 'in': 'body', 'name': 'petId', 'schema': {'format': 'binary', 'type': 'string'}}

    def test_body_binary_octet_stream(empty_swagger_spec, param_spec):
        param_spec['in'] = 'body'
        param_spec['schema'] = {
            'type': 'string',
            'format': 'binary'
        }
        del param_spec['type']
        del param_spec['format']
        request = {
            'headers': {
                'Content-Type': 'application/octet-stream'
            }
        }
        marshal_param(param, bytes(10), request)
>       assert b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' == request['data']
E       assert b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' == '"\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000\\u0000"'

tests/param/marshal_param_test.py:175: AssertionError
```

at the end showing also:
```
src/bravado-core/bravado_core/spec.py:251: Warning: binary format is not registered with bravado-core!
  .format(name), Warning)
```

lucafavatella:
I see two issues:

* bravado-core [does not honour](https://github.com/Yelp/bravado-core/blob/b192c199c7e7303a03ce0e07e11c5bfec25e9d2b/bravado_core/formatter.py#L92-L98) the [OpenAPI spec when it comes to bytes and binary formats](https://github.com/OAI/OpenAPI-Specification/blob/d3b582593a1f6092c8666ad7c893ae5e12779ec7/versions/2.0.md#data-types) - i.e. bytes is misinterpreted (it shall be base64) and binary is ignored;

* [jsonschema ignores bytesequenses in python 3 at all](https://github.com/Julian/jsonschema/blob/7ccd4ff376f8417a71673b7c761e65a9ee8a0bef/jsonschema/validators.py#L56-L60) i.e. the bytes type.

lucafavatella:
> jsonschema ignores bytesequenses in python 3 at all i.e. the bytes python3 type.

jsonschema appears [not running python3 tests on Travis CI](https://github.com/Julian/jsonschema/pull/319#issuecomment-277799344) but willing to accept patches for python3 support. Maybe a minor patch adding bytes among str_types can be written but a test in https://github.com/Julian/jsonschema/blob/master/jsonschema/tests/test_validators.py shall be written.

macisamuele:
@lucafavatella nice finding ...
Checking the [JSON Schema specs](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#page-18) seems that binary and base64 are not valid JSON Schema formats. So we should take care of them within the ``bravado_core`` implementation.
Since you've identified the discrepancy between the defined formats, I would suggest you to fix the ``byte`` format and to implement the ``binary`` Swagger format in formatter.py. If you do so please remember to update the CHANGELOG file writing that the change could not be back-compatible since someone could be using the current "byte" format as "binary".

lucafavatella:
@macisamuele I opened issue tracking those two items. Thinking more about it, it looks like what I want (binary body in request) cannot be achieved in OpenAPI 2.0 by string of format binary because string would still need opening and closing double quotes (`"`) and byte sequence - for as much as arbitrary it is - requires some escaping. So I think what I want more is request body of type file with arbitrary content-type header - apparently not supported in OpenAPI 2.0 (maybe in OpenAPI 3.0 that is work in progress I understand).

lucafavatella:
So closing this ticket.

lucafavatella:
The main OpenAPI ticket appears to be https://github.com/OAI/OpenAPI-Specification/pull/670#issue-150739574

aronatkins:
I have been able to test POSTing a binary file (a `.tar.gz` in my example) in a request with Swagger v2.0.

Swagger YAML snippet:

```yaml
paths:
  /upload:
      parameters:
        - name: archive
          in: body
          schema:
            type: string
            format: binary # byte
          required: true
          description: |
            A `gzip` compressed `tar` archive file.
```

I have monkey-patched `marshal_param` as follows:

```python
import bravado_core.param
_old_marshal_param = bravado_core.param.marshal_param
def _marshal_param(param, value, request):
    swagger_spec = param.swagger_spec
    deref = swagger_spec.deref
    param_spec = deref(bravado_core.param.get_param_type_spec(param))
    location = param.location

    # Binary data is given to the request as-is. This might be a stream or a
    # full array of bytes.
    if location == 'body':
        param_type = param_spec.get('type')
        param_format = param_spec.get('format')
        if param_type == 'string' :
            if param_format in ('byte','binary'):
                request['data'] = value
                return
    
    _old_marshal_param(param, value, request)
bravado_core.param.marshal_param = _marshal_param
```

Folks talk about using `format: binary` or `format: byte`;  this patch allows both.

In test, I can:

```python
archive = io.BytesIO()
with tarfile.open(fileobj=archive, mode='w:gz') as tar:
    content = 'this is the data file'.encode('utf-8')
    info = tarfile.TarInfo('data.txt')
    info.size = len(content)
    tar.addfile(info, io.BytesIO(content))
archive.seek(0) # send a stream that starts at the start.
uploaded = client.archives.upload(archive=archive).result()
```

PR #147, opened 2017-02-07
Fix Python 2.6 build issue

macisamuele:
I'm limiting the ``jsonschema`` dependency in case of Python 2.6 since from [``jsonschema>=2.6.0``](https://github.com/Julian/jsonschema/blob/master/CHANGELOG.rst#v260) it is dropped support for Python 2.6. This solution is a temporary fix until we decide to completely drop Python 2.6 support on ``bravado-core``.
In the mean time I have update ``pre-commit`` hooks and dependencies and added travis tests against python 3.5 and 3.6

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,23 +1,22 @@
-
-- repo: git://github.com/pre-commit/pre-commit-hooks
-  sha: c8a1c91
-  hooks:
-    - id: autopep8-wrapper
-      args: ['-i', '--ignore=E309,E501']
-    - id: check-json
-    - id: check-yaml
-    - id: debug-statements
-    - id: end-of-file-fixer
-    - id: flake8
-      args: [--max-line-length=131]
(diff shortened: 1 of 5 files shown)


macisamuele:
@sjaensch could you have a look to this? As soon as we fix travis we can merge other PRs.
Note: this PR tries to limit issues with Python 2.6 ... do you mind if we remove support for python 2.6? ([bravado](https://github.com/Yelp/bravado/) already dropped support and [swagger_spec_validator PR 64](https://github.com/Yelp/swagger_spec_validator/pull/64) will remove support)

sjaensch:
@maci I'd say drop Python 2.6 instead of adding a package restriction. I've mentioned this in a comment to another pull request - I don't think there is much value in maintaining Python 2.6 support, like you said. Go ahead and get rid of it! :)

macisamuele:
Support for Python 2.6 removed!

sjaensch:
🛳

PR #148, opened 2017-02-07
make unmarshal_collection_format unmarshal empty lists properly

firba1:
I added a test to check this as well. This fixes an issue where if empty string is an invalid value of the array's format, then it doesn't 400. Not to mention, getting `['']` instead of `[]` is just wrong.


macisamuele:
@firba1 Thanks a lot for your contribution. As soon as we build the travis build we'll merge your change.

Additionally will be great if you can fill also the PR information in our CHANGELOG as described on [Contribution](https://github.com/Yelp/bravado-core#contributing) section of the readme.

Issue #150, opened 2017-02-16
partial objects (missing attributes)


dpopowich:
If I have a definition like this:

```json
   "definitions": {
      "Foo": {
         "properties": {
            "id": {
               "type": "integer"
            },
            "name": {
               "type": "string"
            },
            "job": {
               "type": "string"
            }
         },
         "type": "object"
      },
     ...
```

And I reference this in a PUT spec:

```json
"paths": {
      "/foo/{id}": {
         "put": {
            "parameters": [
               ...
               {
                  "required": true,
                  "in": "body",
                  "name": "body",
                  "schema": {
                     "$ref": "#/definitions/Foo"
                  }

               }
            ], ...
```

When I call `PUT /foo/14` with a JSON body of:

```json
{
   "id": 14,
   "name": "joeschmoe"
}
```

The unmarshalled python output from bravado-core is:

```python
{
   "id": 14,
   "name": "joeschmoe",
   "job": None
}
```
That is, `job` was missing, but bravado-core [explicitly looks](https://github.com/Yelp/bravado-core/blob/master/bravado_core/unmarshal.py#L138-L142) for missing properties and sets them on the object with a `None` value.

I cannot find this behavior specified in the swagger spec anywhere. If it exists, can someone point me to this requirement (forcing missing properties to null) in the spec?

If this behavior is not required by the swagger spec and bravado-core is overreaching here, how about an option to allow partial objects?

Otherwise, how is someone supposed to provide PATCH or similar functionality?

sjaensch:
`None` means "unknown" / "not set". The current behavior is to make sure all attributes defined in the swagger spec are present in the unmarshalled object. You may want to discard the attributes that have the value None.

dpopowich:
How do I determine the difference between "value not sent" and "set this value to `None`"?

Or, another way to ask the question, how do I implement PATCH, if bravado-core is going to set every value not delivered to `None`? Imagine an object with 100 properties and I want to send a PATCH to update 5 properties. One can't simply filter out all properties with value `None` because a legal value for a property may be `None` (e.g, one object references another and you want to disassociate the foreign reference so you pass `null` as the value).

Or yet another way to ask the question, where in the swagger spec does it say optional properties are to be treated as null? 

I think bravado-core is overreaching with this behavior. It should at least be configurable.

sjaensch:
[...]

PR #152, opened 2017-02-19
Add nonefy_missing_properties configuration

macisamuele:
This CR target Issue #150 (open by @dpopowich).

I have added an additional configuration flag ``nonefy_missing_properties`` which is defaulted to ``True`` for backward compatibility.

Setting the flag to ``False`` [bravado-core](https://github.com/Yelp/bravado-core/blob/master/bravado_core/unmarshal.py#L138-L142) will not set the to all the missing properties to ``None``.

**NOTE**: ``unmarshal_model`` will not be affected by this change since [``model_constructor``](https://github.com/Yelp/bravado-core/blob/master/bravado_core/model.py#L167-L171) will default to ``None`` all the missing properties.

An example is available on https://gist.github.com/macisamuele/b903295f1935ab7270e31145601dd5cb

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -56,7 +56,10 @@ CONFIG_DEFAULTS = {
     # :class:`bravado_core.formatter.SwaggerFormat`. These formats are in
     # addition to the formats already supported by the Swagger 2.0
     # Specification.
-    'formats': []
+    'formats': [],
+
+    # Fill with None all the missing properties during object unmarshal-ing
+    'include_missing_properties': True,
 }
 
 
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -17,21 +17,25 @@ All configuration is stored in a ``dict``.
     swagger_spec = Spec.from_dict(spec_dict, config=config)
 
 
-========================= =============== =========  ====================================================
-Config key                Type            Default    Description
-------------------------- --------------- ---------  ----------------------------------------------------
-*validate_swagger_spec*   boolean         True       | Validate the Swagger spec against
-                                                     | the Swagger 2.0 Specification.
-------------------------- --------------- ---------  ----------------------------------------------------
-*validate_requests*       boolean         True       | On the client side, validates outgoing requests.
-                                                     | On the server side, validates incoming requests.
-------------------------- --------------- ---------  ----------------------------------------------------
-*validate_responses*      boolean         True       | On the client side, validates incoming responses.
-                                                     | On the server side, validates outgoing responses.
-------------------------- --------------- ---------  ----------------------------------------------------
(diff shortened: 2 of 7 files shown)


sjaensch:
I'd like us to first conclude the discussion in #150. If bravado-core already supports that use case then I'd prefer not to add additional code and even a config option.

macisamuele:
@sjaensch sure ... let's see how the discussion will proceed!
If we figure out a way to avoid the additional config it will be great otherwise we could have a way to support the additional use case without (I hope) bad impacts over the services that are currently dependent on ``bravado-core``.

bravado_core/spec.py
```diff
@@ -51,7 +56,10 @@
     # :class:`bravado_core.formatter.SwaggerFormat`. These formats are in
     # addition to the formats already supported by the Swagger 2.0
     # Specification.
-    'formats': []
+    'formats': [],
+
+    # Fill with None all the missing properties during object unmarshal-ing
+    'nonefy_missing_properties': True,
```
sjaensch: How about `include_missing_properties`? Or something else? I'm not sure people will understand what "Nonefy"  means. :)

docs/source/config.rst
```diff
@@ -17,21 +17,25 @@ All configuration is stored in a ``dict``.
...
+*validate_responses*          boolean         True      | On the client side, validates incoming responses.
+                                                        | On the server side, validates outgoing responses.
+----------------------------- --------------- --------- ----------------------------------------------------
+*use_models*                  boolean         True      | Use python classes to represent models
+                                                        | instead of dicts. See :ref:`models`.
+----------------------------- --------------- --------- ----------------------------------------------------
+*formats*                     list of         []        | List of user-defined formats to support.
+                              SwaggerFormat             | See :ref:`formats`.
+----------------------------- --------------- --------- ----------------------------------------------------
+*nonefy_missing_properties*   boolean         True      | Fill with None properties during
```
sjaensch: "Create properties with the value None if they were not submitted during object unmarshalling"

or something similar, i.e. mostly providing a bit more context.

macisamuele:
Applied @sjaensch naming suggestions

macisamuele:
I have modified ``test_definitions_with_ref`` in order to be executed with all the ``include_missing_properties`` values (``True`` and ``False``).

I would make more evident that using ``bravado-core`` models we are including the missing properties because it will be harder to discriminate if the property is present or not (constraint enforced by [model.py](https://github.com/Yelp/bravado-core/blob/master/bravado_core/model.py#L171)), different from what happen during object unmarshaling (which returns a dictionary).

macisamuele:
* Rebased the branch on top of the new origin/master
* ``model``s do NOT set properties if they're are missing and ``include_missing_properties is False``

PR #154, opened 2017-02-24
Consider default when unmarshalling

### LINKED ISSUE #153: missing properties are not set to default when unmarshalled
If I have an object which I specify as a `"in": "body"` parameter for a PUT or POST with properties that have `default` set, e.g:

```json
"parameters": [
   {
      "required": true,
      "in": "body",
      "name": "body",
      "schema": {
         "type": "object",
         "properties": {
            "name": {
               "type": "string",
            },
            "count": {
               "type": "integer",
               "default": 0
            }
         }
      }
   }, ...
]
```

The default is ignored for missing parameters, e.g, with the following JSON body request:

```json
{
   "name": "Phred"
}
```

The output is:

```python
{
   "name": "Phred",
   "count": None
}
```

I'd expect it to be:

```python
{
   "name": "Phred",
   "count": 0
}
```

Note: `x-nullable` has no effect here: `default` is ignored regardless of the `x-nullable` setting when the property is missing and the value is set to `None`.

sjaensch:
Make sure we consider the default value when unmarshalling. Fixes #153.

### DIFF
--- a/bravado_core/schema.py
+++ b/bravado_core/schema.py
@@ -118,11 +118,15 @@ def handle_null_value(swagger_spec, schema_object_spec):
 
     :param swagger_spec: :class:`bravado_core.spec.Spec`
     :param schema_object_spec: dict
-    :return: None if the spec is nullable
-    :raises: SwaggerMappingError if the spec is not nullable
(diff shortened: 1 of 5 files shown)


macisamuele:
🚢

PR #155, opened 2017-02-26
Update pre-commit, remove Python 3.4 from tests

sjaensch:
- make sure pre-commit hooks are installed whenever people run tests
- also ran fix-encoding-pragma and reorder-python-imports on all files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ docs:
 install:
 	pip install .
 
-test:
+test: install-hooks
 	tox
 
 tests: test
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 -   repo: git://github.com/pre-commit/pre-commit-hooks
-    sha: 46251c9523506b68419aefdf5ff6ff2fbc4506a4
+    sha: a11d9314b22d8f8c7556443875b731ef05965464
     hooks:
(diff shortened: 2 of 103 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ docs:
 install:
 	pip install .
 
-test:
+test: install-hooks
```
macisamuele: Is this needed? ``tox`` will install all the needed dependencies.
Maybe you like to run ``pre-commit`` checks during tests (ie. ``pre-commit run flake8 --all-files``)

sjaensch: flake8 already runs as part of `make test`. The problem with running pre-commit during make test is the virtualenv setup; I'm sure there's a solution, I just didn't want to change that as well in this PR.
install is a separate target and has nothing to do with install-hooks. What's your concern here?

macisamuele: The concern was related to adding a new dependency on the make target that actually is not need; but according to your explanation it seems needed! So just drop what I said 😄

sjaensch: Yeah the way it works for bravado-core is that it installs pre-commit in a separate virtualenv. Moving it to install it during the normal test run (in the same virtualenv) and then running flake8 causes ist to install and run three times (once for each Python version we test).

macisamuele: Thanks for the more detailed explaination.
🚢

PR #159, opened 2017-03-10
Have models inherit from base class, support model polymorphism (continuation of  #128)

macisamuele:
This CR enrich the great modification provided by @jlumpe in #128.
* I have merged master branch solved conflicts and re-added ``include_missing_properties`` configuration
* Added tests for discriminator feature.


CHANGELOG.rst
```diff
@@ -5,6 +5,8 @@ X.X.X (XXXX-XX-XX)
 ------------------
 - Add include_missing_properties configuration - PR #152
 - Consider default when unmarshalling - PR #154
+- Refactor of models generation - PR #128, #159. Thanks Michael Jared Lumpe for your contribution
```
sjaensch: Not sure this requires its own entry in the changelog. Would the next line be enough? Maybe add it in there if you want to mention it?

bravado_core/schema.py
```diff
@@ -66,6 +66,56 @@ def is_list_like(spec):
...
+            found_types.add(composed_type)
+        else:
+            # Fail if we can't determine this unambiguously
+            return None
+
+    if len(found_types) == 1:
+        # Found a single type
+        return found_types.pop()
+
+    if found_types == {'integer', 'number'}:
```
sjaensch: I'm not sure this is correct. Doesn't this mean what we have here is an object composed of two properties, both of which are integers? I.e. in almost every "allOf" case the type should be "object", right?

bravado_core/swagger20_validator.py
```diff
@@ -104,6 +106,59 @@ def enum_validator(swagger_spec, validator, enums, instance, schema):
...
+    discriminator_value = instance[discriminator_attribute]
+    if discriminator_value not in swagger_spec.spec_dict['definitions']:
+        raise ValidationError(
+            message='\'{}\' is not a recognized schema'.format(discriminator_value)
+        )
+
+    new_schema = deepcopy(swagger_spec.deref(swagger_spec.spec_dict['definitions'][discriminator_value]))
+    if 'allOf' not in new_schema:
+        raise ValidationError(
+            message='discriminated schema \'{}\' must inherits from \'{}\''.format(
```
sjaensch: must inherit from

bravado_core/swagger20_validator.py
```diff
@@ -104,6 +106,59 @@ def enum_validator(swagger_spec, validator, enums, instance, schema):
...
+                discriminator_value, schema['x-model']
+            )
+        )
+
+    schemas_to_remove = [s for s in new_schema['allOf'] if is_ref(s) and swagger_spec.deref(s) == schema]
+    if not schemas_to_remove:
+        # Not checking against len(schemas_to_remove) > 1 because it should be prevented by swagger spec validation
+        raise ValidationError(
+            message='discriminated schema \'{}\' must inherits from \'{}\''.format(
```
sjaensch: same here

bravado_core/swagger20_validator.py
```diff
@@ -104,6 +106,59 @@ def enum_validator(swagger_spec, validator, enums, instance, schema):
...
+    """
+
+    discriminator_value = instance[discriminator_attribute]
+    if discriminator_value not in swagger_spec.spec_dict['definitions']:
+        raise ValidationError(
+            message='\'{}\' is not a recognized schema'.format(discriminator_value)
+        )
+
+    new_schema = deepcopy(swagger_spec.deref(swagger_spec.spec_dict['definitions'][discriminator_value]))
+    if 'allOf' not in new_schema:
```
sjaensch: Are you sure this is always correct? What if you have an instance of the base object?

bravado_core/swagger20_validator.py
```diff
@@ -104,6 +106,59 @@ def enum_validator(swagger_spec, validator, enums, instance, schema):
         yield error
 
 
+def discriminator_validator(swagger_spec, validator, discriminator_attribute, instance, schema):
```
sjaensch: This method needs unit tests. :)

bravado_core/schema.py
sjaensch: We should also add a test for the allOf case here.

bravado_core/swagger20_validator.py
macisamuele: This is already tested by [``test_validate_valid_polymorphic_object``](https://github.com/macisamuele/bravado-core/blob/models/tests/validate/validate_object_test.py#L339) and [``test_validate_invalid_polymorphic_object``](https://github.com/macisamuele/bravado-core/blob/models/tests/validate/validate_object_test.py#L401)

bravado_core/swagger20_validator.py
macisamuele: You're right ... I'm adding a test for this case

macisamuele: done

CHANGELOG.rst
macisamuele: Removed it!

bravado_core/schema.py
macisamuele: You're completely right ... as far as I'm understanding in case of ``allOf`` the type should be set to ``object`` (according to [Swagger Specs > Composition and Inheritance (Polymorphism)](http://swagger.io/specification/)).

I'm modifying it and adding the needed tests

bravado_core/model.py
```diff
@@ -69,142 +67,321 @@ def collect_models(container, key, path, models, swagger_spec):
...
+
+    @property
+    def _additional_props(self):
+        """Names of properties in instance which are not defined in spec."""
+        return set(self.__dict).difference(self._properties)
+
+    def _as_dict(self, additional=False, recursive=False):
+        """Get property values as dictionary.
+
+        :param bool additional: Whether to include additional properties
```
sjaensch: I'd name this either additional_props (see above) or additional_properties.

bravado_core/model.py
```diff
@@ -69,142 +67,321 @@ def collect_models(container, key, path, models, swagger_spec):
...
+            if attr_name in self
+        ]
+        return "{0}({1})".format(self.__class__.__name__, ', '.join(s))
+
+    @property
+    def _additional_props(self):
+        """Names of properties in instance which are not defined in spec."""
+        return set(self.__dict).difference(self._properties)
+
+    def _as_dict(self, additional=False, recursive=False):
```
sjaensch: I would assume that `recursive=True` is the expected behavior, can we change that?

tests/model/model_test.py
```diff
@@ -0,0 +1,84 @@
...
+        cat_swagger_spec.spec_dict['definitions']['Cat'], cat_swagger_spec
+    ).keys())
+
+
+def test_model_delete_property(definitions_spec, user_type, user_kwargs):
+    user = user_type(**user_kwargs)
+
+    assert 'id' in set(user)
+    del user.id
+    assert 'id' in set(user)
```
sjaensch: Can you explain a bit what we're testing here? It's not easy to understand if you don't read `model.py` at the same time, especially since you don't assert on the user.id value before and after. I.e. maybe say "deleting a property defined in the spec should set the value to `None`, and assert on the value before and after.

tests/model/model_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+    assert any(
+        user[property] == user_kwargs.get(property)
+        for property in definitions_spec['User']['properties'].keys()
+        if property != 'id'
+    )
```
sjaensch: What is this asserting exactly? The any() makes it really muddy. It says that at least one of the user_kwargs values needs to be equal to the one in the user dict...?

tests/model/model_test.py
```diff
@@ -0,0 +1,84 @@
...
+        del user.not_existing_property
+    assert ex.type == AttributeError
+    assert ex.value.args[0] == 'not_existing_property'
+
+
+def test_model_delete_additional_property(definitions_spec, user_type, user_kwargs):
+    user_type._model_spec['additionalProperties'] = True
+    user = user_type(foo='bar', **user_kwargs)
+
+    assert 'foo' in set(user)
```
sjaensch: I don't think you need the set()...? If you do then maybe it's better to just use user.keys()?

bravado_core/model.py
macisamuele: sure

tests/model/model_test.py
macisamuele: 👍  adding the comment

tests/model/model_test.py
macisamuele: wrong function ... it should be ``all``.
I'm checking that each property (except ``id``) has the same value either in the model (``user``) and in the input dictionary (``user_kwargs``)

tests/model/model_test.py
macisamuele: removing ``set`` and ``.keys()`` since not needed

macisamuele:
* Changed ``additional_properties`` and ``recursive`` default values
* Added tests for ``_as_dict`` with and without recursion
* Updated ``_as_dict`` in order to go recursively through lists

tests/model/model_test.py
```diff
@@ -70,6 +74,43 @@ def test_model_as_dict(definitions_spec, user_type, user_kwargs):
 
 
 @pytest.mark.parametrize(
+    'recursive',
+    [
+        True,
+        False,
+    ]
+)
+def test_marsshal_as_dict_recursive(polymorphic_spec, recursive):
```
sjaensch: marshal

Issue #160, opened 2017-03-21
type object "SomeBravadoResourceType" has no attribute 'marshal' ( and 'unmarshal')


adamhadani:
Started seeing these errors today after installing the just release 4.7.0 bravado-core. Downgrading to previous version (4.6.1) makes these go away so seems like some kind of regression.. Here's an example snippet:

```python
...
chatroom_event.marshal(),
  File "/Users/adamhadani/.virtualenvs/venv/lib/python3.6/site-packages/bravado_core/model.py", line 176, in __getattr__
    .format(type(self).__name__, attr_name)
AttributeError: type object 'SomeBravadoResourceType' has no attribute 'marshal'
```

Also, seeing similar error when trying to use `unmarshal`:

```python
...
.unmarshal({
AttributeError: type object 'SomeBravadoResourceType' has no attribute 'unmarshal'
```

adamhadani:
FYI @sjaensch

macisamuele:
@adamhadani you're right ... version 4.7.0 changes the way to marshal and unmarshal models.
Instead of using ``marshal`` and ``unmarshal`` class methods are replaced by ``_marshal`` and ``_unmarshal`` ones. This was injected by #159.

Since actually this is a back incompatible change we should change the package version and make evident in the changelog how to migrate from ``v4.6.1`` to ``v5.0.0`` (or any version code that will be attributed to the new version)

Thanks a lot for highlighting this ... we'll try to fix this regression soon!

adamhadani:
ping @macisamuele I have a PR  on the way for this which re-introduces original function names with a deprecation warning, will ping u there when up, thanks for confirmation!

sjaensch:
Fixed by #161.

PR #161, opened 2017-03-21
Add back old private method names with deprecation warning

adamhadani:
See related Issue #160 for fuller description of underlying issue.
TL;DR - there was an API change that is backward incompatible which is breaking usage of the library. This change re-introduces the old method names with a deprecation warning to allow for a migration grace period.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 - The content will be used to build the Changelog for the new bravado-core release
   (add before this line your modifications summary and PR reference)
 
+4.7.1 (2017-03-21)
+------------------
+- Fix backward-incompatible Model API change which renames all model methods to have a single underscore infront of them. A deprecation warning has been added.
+
 4.7.0 (2017-03-21)
 ------------------
 - Added support for nullable fields in the format validator - PR #143. Thanks Adam Ever-Hadani
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 - The content will be used to build the Changelog for the new bravado-core release
   (add before this line your modifications summary and PR reference)
 
-4.7.1 (2017-03-21)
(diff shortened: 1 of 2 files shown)


adamhadani:
ping @macisamuele note I didn't re-add all of the methods potentially - This fixes the issues we personally ran into which probably reflect on most people's usage, but you may want to augment it with similar mechanism for other private methods that may have changed

bravado_core/model.py
```diff
@@ -306,6 +307,13 @@ def _from_dict(cls, dct):
         )
         return model
 
+    def marshal(self):
+        warn(
+            "Private model object methods are now prefixed with single underscore - use _marshal() instead.",
```
sjaensch: They're not private, they just use a single underscore to avoid name clashes with attributes (which are coming from the swagger spec). This is unfortunate, but a common practice in Python when objects have dynamic attributes. See namedtuple for an example.

CHANGELOG.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 - The content will be used to build the Changelog for the new bravado-core release
   (add before this line your modifications summary and PR reference)
 
+4.7.1 (2017-03-21)
```
sjaensch: 2017-03-22 please. :)

bravado_core/model.py
```diff
@@ -315,6 +323,14 @@ def _marshal(self):
         return marshal_model(self._swagger_spec, self._model_spec, self)
 
     @classmethod
+    def unmarshal(cls, val):
```
sjaensch: Could you maybe copy tests.model.create_model_type_test.test_marshal_and_unmarshal and use the deprecated methods just to make sure they work? We can remove the test once we remove these methods.

sjaensch:
I've decided to make the requested changes myself in order to expedite the release, thanks for your contribution @adamhadani!

adamhadani:
Thanks Stephan!

Issue #163, opened 2017-03-22
unmarshal bug in 4.7.(0|1)


CBarraford:
I've confirmed that we don't get the following error when i downgrade from 4.7.0 (or 4.7.1) to 4.6.1.

```
======================================================================
1) ERROR: test_bundles (bundles.test_bundles.BundlesTest)
----------------------------------------------------------------------
   Traceback (most recent call last):
    test/api/bundles/test_bundles.py line 30 in test_bundles
      app = self.client.Applications.Create(body=dict(name='test_app')).result()
    /usr/local/lib/python2.7/dist-packages/bravado/http_future.py line 77 in result
      self.response_callbacks)
    /usr/local/lib/python2.7/dist-packages/bravado/http_future.py line 115 in unmarshal_response
      operation)
    /usr/local/lib/python2.7/dist-packages/bravado_core/response.py line 112 in unmarshal_response
      op.swagger_spec, content_spec, content_value)
    /usr/local/lib/python2.7/dist-packages/bravado_core/unmarshal.py line 56 in unmarshal_schema_object
      return unmarshal_model(swagger_spec, schema_object_spec, value)
    /usr/local/lib/python2.7/dist-packages/bravado_core/unmarshal.py line 194 in unmarshal_model
      model_as_dict = unmarshal_object(swagger_spec, model_spec, model_value)
    /usr/local/lib/python2.7/dist-packages/bravado_core/unmarshal.py line 136 in unmarshal_object
      if schema.has_default(swagger_spec, prop_spec):
    /usr/local/lib/python2.7/dist-packages/bravado_core/schema.py line 22 in has_default
      return 'default' in swagger_spec.deref(schema_object_spec)
   TypeError: argument of type 'NoneType' is not iterable

```

CBarraford:
My API responded with 200, here's an example JSON body
```
{
  "id": 12,
  "principal_id": 1,
  "access_type": "acl",
  "connection_timeout": {
    "Int64": 0,
    "Valid": false
  },
  "read_timeout_timeout": {
    "Int64": 0,
    "Valid": false
  },
  "init_timeout": {
    "Int64": 0,
    "Valid": false
  },
  "idle_timeout": {
    "Int64": 0,
    "Valid": false
  },
  "max_processes": {
    "Int64": 0,
    "Valid": false
  },
  "min_processes": {
    "Int64": 0,
    "Valid": false
  },
  "max_conns_per_process": {
    "Int64": 0,
    "Valid": false
  },
[...]

PR #164, opened 2017-03-22
always return a dict for get_spec_for_prop

CBarraford:
Fixes https://github.com/Yelp/bravado-core/issues/163
The function doc for `get_spec_for_prop` says "rtype:dict", but there is a case where a `None` is returned which is not handled properly downstream.

### DIFF
--- a/bravado_core/schema.py
+++ b/bravado_core/schema.py
@@ -100,7 +100,7 @@ def get_spec_for_prop(swagger_spec, object_spec, object_value, prop_name):
     if isinstance(additional_props, bool):
         # no spec for additional properties to conform to - this is basically
         # a way to send pretty much anything across the wire as is.
-        return None
+        return dict()
 
     additional_props = deref(additional_props)
     if is_dict_like(additional_props):


sjaensch:
I don't think this is the correct way to fix it, rather we should make sure that `unmarshal_object()` passes through the data unmodified when no spec is available. By adding handling for default values we unintentionally broke this case for properties without a spec and a `null` value.

So what I'd do would be to modify the if condition to say `if v is None and k not in required_fields and prop_spec:` and then add a test that makes sure we do the right thing. `test_pass_through_property_with_no_spec` is a good starting point.

CBarraford:
@sjaensch im ok with doing it that way, but i would say maybe the doc for the func should be updated so that its return value is a dict or none?

sjaensch:
@CBarraford absolutely, good idea! Do you want to do it? You could have modified your PR accordingly. :) Otherwise I have a branch ready I can publish.

CBarraford:
@sjaensch if you got a branch ready to publish 👍

sjaensch:
Done, #165.

PR #167, opened 2017-04-05
Fix marshall and validation of allOf

plumzl:
Support composition when object is passed in request body.


plumzl:
Can I have a review on this change?

sjaensch:
I've just released version 4.7.3 that contains your fix. Thanks again for your contribution @plumzl!

PR #171, opened 2017-04-19
Change how we handle defaults.

jnb:
* Only allow a None to be returned for an optional field if that field is x-nullable.
* Instantiate a missing field to its default value if provided.

### DIFF
--- a/bravado_core/unmarshal.py
+++ b/bravado_core/unmarshal.py
@@ -147,6 +147,8 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
     for prop_name, prop_spec in iteritems(properties):
         if prop_name not in result and swagger_spec.config['include_missing_properties']:
             result[prop_name] = None
+            if schema.has_default(swagger_spec, prop_spec):
+                result[prop_name] = schema.get_default(swagger_spec, prop_spec)
 
     return result
 
--- a/tests/unmarshal/unmarshal_object_test.py
+++ b/tests/unmarshal/unmarshal_object_test.py
@@ -127,6 +127,17 @@ def test_with_properties(empty_swagger_spec, address_spec, address, street_type,
     assert expected_address == result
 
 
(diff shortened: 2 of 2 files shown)


bravado_core/unmarshal.py
```diff
@@ -126,18 +126,12 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
             type(object_value), object_value))
 
     object_spec = deref(object_spec)
-    required_fields = object_spec.get('required', [])
 
     result = {}
     for k, v in iteritems(object_value):
         prop_spec = get_spec_for_prop(
             swagger_spec, object_spec, object_value, k)
-        if v is None and k not in required_fields and prop_spec:
```
jnb: I'd be OK to revert this hunk if you think it's not worth the amount of pain it might cause?

sjaensch: I think the change makes sense conceptually, not sure how many people depend on being allowed to send `None` for optional fields though.

tests/unmarshal/unmarshal_object_test.py
```diff
@@ -459,8 +470,10 @@ def test_non_required_none_value(empty_swagger_spec, property_type):
                                          nullable=False,
                                          property_type=property_type)
     value = {'x': None}
-    result = unmarshal_object(empty_swagger_spec, content_spec, value)
-    assert result == {'x': None}
+    # This should fail because the field is not nullable
+    with pytest.raises(SwaggerMappingError) as excinfo:
+        unmarshal_object(empty_swagger_spec, content_spec, value)
+    assert 'is a required value' in str(excinfo.value)
```
sjaensch: This is a misleading error message, as the field is not required. We just don't allow `None` as value anymore.

jnb:
OK, I scoped this down a bit.  Now the behavior is just to add a default field value in case of a missing field.  If there's isn't a default field specified, then we add a None (as before).

sjaensch:
I've released version 4.7.3 with this change.

PR #175, opened 2017-05-31
Handle responses as JSON only if required to.

mostrows2:
A JSON response will have "application/json", and only if this is set should
the response content be decoded as such.  Otherwise (until something smarter
comes along), on might as well return the raw response content (which
happens to be the correct thing to do for "text/plain" content).

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -101,15 +101,19 @@ def unmarshal_response(response, op):
     if not has_content(response_spec):
         return None
 
+    content_type = response.headers.get('content-type', '').lower()
+
+    if content_type.startswith(APP_JSON):
+        content_spec = deref(response_spec['schema'])
+        content_value = response.json()
+
+        if op.swagger_spec.config['validate_responses']:
+            validate_schema_object(op.swagger_spec, content_spec, content_value)
+
+        return unmarshal_schema_object(
+            op.swagger_spec, content_spec, content_value)
     # TODO: Non-json response contents
-    content_spec = deref(response_spec['schema'])
-    content_value = response.json()
-
-    if op.swagger_spec.config['validate_responses']:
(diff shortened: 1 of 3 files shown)


bravado_core/response.py
```diff
@@ -101,15 +101,19 @@ def has_content(response_spec):
     if not has_content(response_spec):
         return None
 
-    # TODO: Non-json response contents
-    content_spec = deref(response_spec['schema'])
-    content_value = response.json()
+    content_type = response.headers.get('content-type', '').lower()
+
+    if content_type.startswith(APP_JSON):
```
sjaensch: Any reason you went with startswith instead of equality?

mostrows2: To opportunistically capture "application/json-rpc" should that ever occur? Admittedly this test may need to evolve over time since there are various other content types that could contain json. I don't feel strongly about this either way.

sjaensch:
This might be a breaking change for users where the server isn't sending the correct content type header. I'm wondering if we should make this behavior configurable, i.e. add an option to make bravado-core behave the way it used to. The other option would be to ship this as-is and wait for user feedback. Thoughts?

mostrows2:
My approach here is to KISS.  And yes in the future there will be some
breakage. But there's arguably more value in detecting the breakage,
understanding the variety of non-compliant systems and putting in explicit
fixes.  The alternative is that you'd never be sure as to whether the
configuability you've added preemptively is ever used and therefore whether
you could ever change it - since you don't know how/if it is being used.

For that reason I don't think the existing code (which assumed JSON
content) was a bad idea - it was good enough until the time it proved to be
insufficient and is now being updated because of a real need.

--
Michal Ostrowski

sjaensch:
Yeah, that sounds reasonable. Let's make sure we release this as a major version.

PR #177, opened 2017-06-22
Add Flattened specs feature

macisamuele:
With this PR bravado-core is able to perform swagger spec flattening to a single file.
NOTE: ``_marshal_uri`` and ``flattened_spec.descend`` methods are extracted and adapted from [``pyramid_swagger``](https://github.com/striglia/pyramid_swagger/)


tests/spec/Spec/flattened_spec_test.py
```diff
@@ -0,0 +1,173 @@
...
+    flattened_multi_file_recursive_dict, multi_file_recursive_spec,
+):
+    flattened_spec = multi_file_recursive_spec.flattened_spec
+    validator20.validate_spec(
+        # Deep copy needed because validate_spec adds x-scope information
+        spec_dict=copy.deepcopy(flattened_spec),
+        spec_url='',
+        http_handlers={},
+    )
+    # TODO: fix test failure on py27
```
sjaensch: The build is passing, I guess this TODO is done?

tests/formatter/to_wire_test.py
```diff
@@ -9,6 +9,11 @@
 from bravado_core.formatter import to_wire
 from bravado_core.spec import Spec
 
+try:
+    long
+except NameError:
+    from six import integer_types as long
```
sjaensch: This is not the way to do it (and it's not the way it's done further down). I guess this is necessary for flake8 or some other linter that is now using Python 3?

1. You're not using six where you should be (detecting the Python version), and then using six when you don't need to (alias the int type). In fact, you have a bug here since `integer_types` is (as the name suggests) a tuple of types, not a type!
2. Please change all occurences below of `six.PY3` to `not six.PY2` (or use `six.PY2` and switch the if/else if it's in that context). This is better as it will keep doing the right thing if/when Python 4 is released.

tests/spec/Spec/flattened_spec_test.py
macisamuele: Yep ... I forgot to remove it

tests/formatter/to_wire_test.py
macisamuele: I know .. it's a little ugly .. but if we run ``flake8`` using python3 (and soon or later python2 will be decommissioned) we'll get some errors.
```
YELP-MACI:bravado-core maci$ pre-commit run flake8 --all-files
Flake8...................................................................Failed
hookid: flake8

tests/formatter/to_wire_test.py:59:62: F821 undefined name 'long'
tests/formatter/to_wire_test.py:60:16: F821 undefined name 'long'
tests/formatter/to_wire_test.py:61:35: F821 undefined name 'long'
tests/formatter/to_wire_test.py:71:16: F821 undefined name 'long'
tests/formatter/to_wire_test.py:72:35: F821 undefined name 'long'
tests/formatter/to_wire_test.py:79:58: F821 undefined name 'long'
tests/formatter/to_python_test.py:49:64: F821 undefined name 'long'
tests/formatter/to_python_test.py:50:16: F821 undefined name 'long'
tests/formatter/to_python_test.py:60:16: F821 undefined name 'long'
tests/formatter/to_python_test.py:61:35: F821 undefined name 'long'
tests/formatter/to_python_test.py:68:60: F821 undefined name 'long'
```
For this reason I tried to define ``long``, maybe something like
```python
if six.PY3:
    # long definition needed to make happy ``flake8`` on python3
    long = int
```
could be better. Shorter and easier to read.

macisamuele: About the second point I applied the following schema
```python
    if six.PY2:
        ...
    elif six.PY3:
        ...
    else:
        assert False, 'Test not executed with python 2 or python 3'
```

macisamuele: As you asked ... I'll provide an extra PR for pre-commit hooks related changes

sjaensch: Why did you do that (in regards to if/else)? I should have explained better maybe.

It's generally bad practice to test for Python 3 since it means that once Python 4 comes out the Python 2 codepath will be taken (which is almost certainly wrong). That's why you should check for Python 2 and do any special handling there. If you need to do something on Python 3, use `not six.PY2` to make sure it works on Python 4+ as well.

In fact, your else is precisely what I wanted to avoid: break the test on Python 4 although it could very well be the case that it runs fine on that Python version. In summary:

1. Do `if six.PY2: ... else: ...`.
2. Do `if not six.PY2: long = int`

macisamuele: 👍  ... removing the elif for python 3

macisamuele: Fixed on https://github.com/macisamuele/bravado-core/commit/a81e01ecd8b0c332adf6c9764295da32c78e4106

tests/spec/Spec/flattened_spec_test.py
macisamuele: Done

sjaensch:
@josiah-wolf-oberholtzer I've just released version 4.8.0 with this change.

bravado_core/spec_flattening.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+_TYPE_SCHEMA, _TYPE_PATH_ITEM, _TYPE_PARAMETER = range(3)
+
+
+def _determine_object_type(object_dict):
+    """
+    Use best guess to determine the object type based on the object keys.
+    NOTE: it assumes that the base swagger specs are validated and perform type
+    detection for three types of object (only one that could be referenced in the specs):
```
ostacey: The way I'm reading the spec, Swagger also allows for references in a response object. See:

https://swagger.io/specification/#responsesObject

PR #184, opened 2017-08-29
Improve specs flattening: references are allowed for responses

macisamuele:
As reported by @ostacey on #177 
> Swagger also allows for references in a response object.

This PR improves the ``_determine_object_type`` logic in order to recognize also response objects.

The new logic allows to generate ``responses`` 
> An object to hold responses that can be used across operations. This property does not define global responses for all operations.

In addition to that this PR takes care of exposing un-referenced models during flattening process.
This is important especially for Swagger Specs that are taking advantage of polymorphism.
Easier understanding is obtainable checking ``test-data`` diff on [SHA a62d5452f2b54547f1f501d7a447aad8fb3832df](https://github.com/Yelp/bravado-core/pull/184/commits/a62d5452f2b54547f1f501d7a447aad8fb3832df#diff-5b09424ae36a4abaf8109684719fef66)


bravado_core/spec_flattening.py
```diff
@@ -152,24 +160,46 @@ def _determine_object_type(object_dict):
...
+
+            # If description field is specified and there are no other fields other the allowed response fields
+            if 'description' in object_keys and not object_keys - response_allowed_keys:
+                return _TYPE_RESPONSE
+            else:
+                # A schema object has:
+                #  - no mandatory parameters
+                #  - long list of optional parameters (ie. description, type, items, properties, discriminator, etc.)
+                #  - no other fields are allowed
+                # NOTE: In case the method is miss determining the type of a schema object, confusing it with a
```
sjaensch: misdetermining (or mis-determining)

bravado_core/spec_flattening.py
```diff
@@ -152,24 +160,46 @@ def _determine_object_type(object_dict):
...
+                #       response type it will be enough to add, to the object, one key that is not defined
+                #       in ``response_allowed_keys``.  (ie. "additionalProperties": True  (implicitly defined be specs)
```
sjaensch: I thought that the default for additionalProperties was the empty schema (which behaves differently from it being `true`).

bravado_core/spec_flattening.py
```diff
@@ -208,9 +240,15 @@ def flattened_spec(
             handlers=http_handlers or {},
         )
 
+    if spec_definitions is None:
+        warnings.warn(
+            message='Un-referenced models cannot be un-flattened if spec_definitions is not present',
+            category=Warning,
+        )
```
sjaensch: Why is this a warning? Shouldn't this be an information at most? Such models are not used, or did I misunderstand that? Also, should this be "flattened" instead of "un-flattened"?

bravado_core/spec_flattening.py
macisamuele: According to https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.4.1 valid values are:
 - boolean
 - schema object

If the value is ``True`` any object/array/primitive type could be used as additional property.

The default behaviour of bravado-core is to check only the properties defined in the specs, so it means that is implicitly assuming ``additionalProperties`` set to ``True``

bravado_core/spec_flattening.py
macisamuele: 👍

bravado_core/spec_flattening.py
macisamuele: The warning is needed because ``flattened_spec`` is a "public" method and so external applications could already use it. That is also the reason of having ``spec_definitions`` defaulted to ``None``.

Internally to bravado core definitions will always passed so it should never warn.

bravado_core/spec_flattening.py
sjaensch: Why did you stop reading there. :)

    5.4.4.3.  Default values
    [...]
    If "additionalProperties" is absent, it may be considered present
    with an empty schema as a value.

macisamuele: 😄 right ... I'll update the comment and fix the number of brackets too

Issue #185, opened 2017-09-01
x-nullable working only for required properties of objects


janekbaraniewski:
Hey, I've stumbled upon an unexpected behaviour when using `x-nullable` 

My swagger definition:
```
...
        "SomeObject": {
            "type": "object",
            "properties": {
                "data": {
                    "type": "object",
                    "required": [
                        "type"
                    ],
                    "properties": {
                        "type": {
                            "type": "string"
                        },
                        "attributes": {
                            "type": "object",
                            "properties": {
                                "name": {
                                    "type": "string"
                                },
                                "value": {
                                    "type": "string"
                                },
                                "nullable_field": {
                                    "type": "string",
                                    "x-nullable": true
                                }
                            }
                        }
                    }
                }
            }
        }
...
```
From what is stated in [bravado-core documentation](http://bravado-core.readthedocs.io/en/latest/models.html#allowing-null-values-for-properties) I should be able to send in my request `None` as a value of `nullable_field` but sadly that's not the case, `nullable_field` simply doesn't exist in the generated payload. 

I've traced the issue to `marshal_object` method, where if the value of the field is `None` and this field is not required, then the field is simply omitted. This is inconsistent with documentation, where there is no mention that field has to be a required field in order for this to work - even more, in example in documentation field that is being set to null is also not required.

If I'm right and this is a bug I'll be happy to submit PR with a fix, and if not could you explain why it is implemented in this way?

macisamuele:
@janekbaraniewski I quickly checked the implementation of [``handle_null_value``](https://github.com/Yelp/bravado-core/blob/27ed3d16c9ba734ec177ea4c393afba515ba6334/bravado_core/schema.py#L127) and the result that you're getting seems correct according to the code.

From you're prospective this seems wrong and I agree with you. This is a bug.

I think that we have no tests for nullable properties that could be also not-required properties.

I would have no time until Monday/Tuesday to dig more in the code and publish a PR, but if you will be able to do prepare the PR it will be really appreciated.

janekbaraniewski:
hey @macisamuele, thanks for the quick response.
Yeah, sure, I should have PR ready within few hours :)

janekbaraniewski:
PR #186 is ready for review :)

sjaensch:
[...]

PR #186, opened 2017-09-01
fix: working x-nullable for not required properties

janekbaraniewski:
fixes bug which caused not required None values to be omitted regardless of x-nullable being True
issue #185

### DIFF
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -11,6 +11,7 @@ from bravado_core.schema import get_spec_for_prop
 from bravado_core.schema import handle_null_value
 from bravado_core.schema import is_dict_like
 from bravado_core.schema import is_list_like
+from bravado_core.schema import is_prop_nullable
 from bravado_core.schema import SWAGGER_PRIMITIVES
 
 
@@ -142,7 +143,9 @@ def marshal_object(swagger_spec, object_spec, object_value):
             swagger_spec, object_spec, object_value, k)
 
         if v is None and k not in required_fields:
-            continue
+            if not is_prop_nullable(swagger_spec, prop_spec):
+                continue
+
         if prop_spec:
             result[k] = marshal_schema_object(swagger_spec, prop_spec, v)
         else:
--- a/tests/marshal/marshal_object_test.py
+++ b/tests/marshal/marshal_object_test.py
@@ -241,7 +241,7 @@ def test_recursive_ref_with_depth_n(recursive_swagger_spec):
     assert result == expected
 
 
-def test_marshal_with_nullable_property(empty_swagger_spec):
+def test_marshal_with_nullable_required_property(empty_swagger_spec):
     object_spec = {
         'type': 'object',
         'required': ['x'],
@@ -257,7 +257,7 @@ def test_marshal_with_nullable_property(empty_swagger_spec):
     assert result == value
 
 
-def test_marshal_with_non_nullable_property(empty_swagger_spec):
+def test_marshal_with_non_nullable_required_property(empty_swagger_spec):
     object_spec = {
         'type': 'object',
         'required': ['x'],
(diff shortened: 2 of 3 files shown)


bravado_core/marshal.py
```diff
@@ -140,8 +141,11 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
         prop_spec = get_spec_for_prop(
             swagger_spec, object_spec, object_value, k)
-
-        if v is None and k not in required_fields:
+        if (
+            v is None and
+            not is_prop_nullable(swagger_spec, prop_spec) and
+            k not in required_fields
+        ):
             continue
```
macisamuele: Improve code readability I would suggest you to keep the old ``if`` statement and to replace ``continue``  with the extra condition that you need to verify

tests/marshal/marshal_object_test.py
```diff
@@ -273,31 +273,30 @@ def test_marshal_with_non_nullable_property(empty_swagger_spec):
...
         }
     }
     value = {'x': None}
     result = marshal_object(empty_swagger_spec, object_spec, value)
-    assert result == {}
+    assert result == value
 
 
-def test_marshal_with_required_property(empty_swagger_spec):
+def test_marshal_with_non_nullable_non_required_property(empty_swagger_spec):
```
macisamuele: Please split this test in ``test_marshal_with_non_nullable_non_required_property_success`` and ``test_marshal_with_non_nullable_non_required_property_fail`` (or similar naming) and test either the correct marshaling and the incorrect marshaling that will raise ``SwaggerMappingError`` exception

sjaensch:
Heh this is interesting. We've actually never intended for `x-nullable` to be used this way. It was always to allow bravado (as the client) to not complain about `null` values being sent from the server. That's why we have a [battery of tests](https://github.com/Yelp/bravado-core/blob/master/tests/unmarshal/unmarshal_object_test.py#L403) testing every possible case for unmarshalling.

I think it makes sense for marshalling too, so I'm all for it. I see that @macisamuele has just replied as I was writing this. :)

bravado_core/marshal.py
janekbaraniewski: What do you think about something like this:
```python
if v is None and not is_prop_nullable(swagger_spec, prop_spec):
    if k not in required_fields:
        continue
    else:
        raise SwaggerMappingErrror(
            'Required property {0} cannot be None'.format(k))
```
That way conditions that check for `None` are grouped together.

tests/marshal/marshal_object_test.py
janekbaraniewski: I don't think I understand, how would it raise `SwaggerMappingError`?  The result for non nullable non required properties didn't change - they still just won't be marshalled so exception will never be raised.

macisamuele: You're right ... in this condition is not possible to have an error, since the property is optional and not nullable it won't be present on the final object. My bad 😄

bravado_core/marshal.py
macisamuele: I was thinking to
```python
if v is None and k not in required_fields:
    if not is_prop_nullable(swagger_spec, prop_spec):
        continue
```
but I'm fine in you're way of defining with a single logic operation.

Only one small optimization request could you reorder the condition in
I expect that evaluating ``is_prop_nullable(...)`` is more expensive than checking if a key is in a dictionary, could you swap the 2nd with the 3rd logic condition? It won't change the final result (since ``and``ed) but would reduce performance degradation

janekbaraniewski: Yeah, I didn't think about that. Well then I like your version better than mine :) corrected

janekbaraniewski:
Sure thing. Version 4.8.2?

PR #191, opened 2017-09-06
Ensure model discovery runs for flattened specs

### LINKED ISSUE #190: Spec Flattening assumes that all the models were discovered
While building [``pyramid_swagger``](https://travis-ci.org/macisamuele/pyramid_swagger/builds/271496637) I got a list of tests failure related to `` KeyError: 'x-model'`` in [``bravado_core. spec_flattening .flattened_spec``](https://github.com/Yelp/bravado-core/blob/v4.8.3/bravado_core/spec_flattening.py#L303-L306).

Minimal example to reproduce the issue:
- ``swagger.json``
```json
{
  "swagger": "2.0",
  "info": {
    "title": "Test",
    "version": "1.0"
  },
  "paths": {},
  "definitions": {
    "object": {
      "properties": {
        "property": {
          "$ref": "aux.json#/definitions/referenced_object"
        }
      }
    }
  }
}
```
- ``aux.json``
```json
{
    "definitions": {
        "referenced_object": {
            "type": "object"
        }
    }
}
```
- ``test.py``
```python
import json
from os.path import abspath

root_swagger_path = abspath('swagger.json')
with open(root_swagger_path) as f:
    spec_dict = json.load(f)

spec = Spec.from_dict(spec_dict, origin_url='file://{}'.format(root_swagger_path))
spec.flattened_spec
```
Running ``test.py`` generates the following traceback
```
  File "test.py", line 10, in <module>
    spec.flattened_spec
  File "/Users/maci/pg/bravado-core/.tox/py35/lib/python3.5/site-packages/bravado_core/spec.py", line 305, in flattened_spec
    spec_definitions=self.definitions,
  File "/Users/maci/pg/bravado-core/.tox/py35/lib/python3.5/site-packages/bravado_core/spec_flattening.py", line 305, in flattened_spec
    for definition in itervalues(known_mappings['definitions'])
  File "/Users/maci/pg/bravado-core/.tox/py35/lib/python3.5/site-packages/bravado_core/spec_flattening.py", line 305, in <setcomp>
    for definition in itervalues(known_mappings['definitions'])
KeyError: 'x-model'
```

The issue is caused by the wrong assumption that definitions are tagged as model (have ``x-model`` property).

To fix this we could:
1. make sure that ``flattened_spec`` does not assume that ``x-model`` will always be present
2. improve models tagging (currently ``tag_models`` tags only schema objects on ``/definitions`` of the root swagger specs file

NOTE: the second approach requires more time, so I would suggest to go with point 1 first

macisamuele:
This PR is meant to fix #190.

The main difference introduced in the discovery of the missing models that needs to be injected in the flattened specs.

Basically running ``Spec.from_dict`` will trigger the model discovery and tagging process ([Code](https://github.com/Yelp/bravado-core/blob/v4.8.3/bravado_core/spec.py#L170)) which will modify the injected swagger specs dictionary, and so it will add ``x-model`` into ``known_mappings['definitions']`` objects.

In order to reduce as much as possible the overhead of this operation I'm passing to ``Spec.from_dict`` an object that looks like swagger specs (it won't be valid since ``info`` and ``paths`` are missing) and I'm preventing spec validation.
I know that this could not be 100% clear so I hope that the comments will help 😄

### DIFF
--- a/bravado_core/spec_flattening.py
+++ b/bravado_core/spec_flattening.py
@@ -317,9 +317,11 @@ def flattened_spec(
         )
 
         flatten_models = {
-            definition[MODEL_MARKER]
+            # Needed because schema objects could not contain "type" property and so they won't be tagged as models
+            definition.get(MODEL_MARKER)
             for definition in itervalues(known_mappings['definitions'])
         }
+
         for model_name, model_type in iteritems(spec_definitions):
             if model_name in flatten_models:
                 continue
--- a/bravado_core/spec_flattening.py
+++ b/bravado_core/spec_flattening.py
@@ -15,6 +15,7 @@ from six.moves.urllib.parse import urlunparse
 from six.moves.urllib_parse import ParseResult
 from swagger_spec_validator.ref_validators import in_scope
 
+from bravado_core.model import MODEL_MARKER
 from bravado_core.schema import is_dict_like
 from bravado_core.schema import is_list_like
 from bravado_core.schema import is_ref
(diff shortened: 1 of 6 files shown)


macisamuele:
@sjaensch before merge this I'm checking that this is not causing other ``pyramid_swagger`` failures.

macisamuele:
I had to update this PR because ``pyramid_swagger`` build was still failing [travis](https://travis-ci.org/macisamuele/pyramid_swagger/builds/272412418).
The new change let ``pyramid_swagger`` to have green tests (at least not failing due to ``bravado-core``) [travis](https://travis-ci.org/macisamuele/pyramid_swagger/builds/272442524)

bravado_core/spec_flattening.py
```diff
@@ -317,9 +317,11 @@ def descend(value):
         )
 
         flatten_models = {
-            definition[MODEL_MARKER]
+            # Needed because schema objects could not contain "type" property and so they won't be tagged as models
```
sjaensch: How about this?

````python
# schema objects might not have a "type" set so they won't be tagged as models

macisamuele: I like it more 😄

PR #195, opened 2017-09-26
Reintroduce instance check while marshaling models

macisamuele:
#194 removed the check around Model class while marshaling a model.

This PR:
 * reintroduces the check
 * makes sure that a model instance will be recognized as a valid model class even if the model type is originated by a different ``bravado_core.spec.Spec`` object build on top of the same swagger specs.
 * deprecates usage of ``_isinstance`` instance method suggesting the use of the *standard* ``isinstance`` python method

### DIFF
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -176,10 +176,11 @@ def marshal_model(swagger_spec, model_spec, model_value):
         return handle_null_value(swagger_spec, model_spec)
     model_type._isinstance(model_value)
     if not isinstance(model_value, model_type):
-        # if not model_type._isinstance(model_value):
         raise SwaggerMappingError(
-            'Expected model of type {0} for {1}:{2}'
-            .format(model_name, model_value.__class__.__name__, model_value))
+            'Expected model of type {0} but got {1} for value {2}'.format(
+                model_name, model_value.__class__.__name__, model_value,
+            ),
+        )
 
     # just convert the model to a dict and feed into `marshal_object` because
     # models are essentially 'type':'object' when marshaled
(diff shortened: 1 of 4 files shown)


bravado_core/marshal.py
```diff
@@ -174,11 +174,12 @@ def marshal_model(swagger_spec, model_spec, model_value):
 
     if model_value is None:
         return handle_null_value(swagger_spec, model_spec)
-
-    if not hasattr(model_value, '_as_dict'):
+    model_type._isinstance(model_value)
+    if not isinstance(model_value, model_type):
+        # if not model_type._isinstance(model_value):
```
sjaensch: I think this line and the one two lines above should be removed, right?

bravado_core/marshal.py
```diff
@@ -174,11 +174,12 @@ def marshal_model(swagger_spec, model_spec, model_value):
...
         raise SwaggerMappingError(
             'Expected model of type {0} for {1}:{2}'
```
sjaensch: How about 'Expected model of type {0} but got {1} for value {2}'

bravado_core/model.py
```diff
@@ -69,6 +71,30 @@ def collect_models(container, key, path, models, swagger_spec):
...
+        """
+        An object is instance of a specific model class if:
+        - model class and instance share the same model Swagger Specs
+        - instance inherits from model class (in case of multiple inheritance, ie. allOff)
+        """
+        if not isinstance(type(instance), type(self)):
+            return False
+
+        if not hasattr(self, '_model_spec'):
+            # This is a generic Model
```
sjaensch: This is only true for the base Model class, right? Or am I missing something? If this is the case then let's be more specific (i.e. "This is the base Model class").

PR #196, opened 2017-09-29
Fixed validate_schema_object to handle None types

mikeecb:
Fixed a small bug would cause swagger validation errors when the obj_type is None.


bplotnick:
👍

PR #200, opened 2017-10-12
feat/Sanitise resource names to ensure they are valid Python identifiers

jakul:
Clean resource names to ensure they:

* can be used as valid Python identifiers, and
* match the naming style guide for the operation sanitisation

This is a breaking change, becuase resources which are currently called `My Tag` will now be called `My_Tag`. For users of `bravado` it will mean workarounds like `getattr(client, 'My Tag').api_method().result()` will need to be replaced with `client.My_Tag.api_method().result()`. I'm not sure what the impact of that change will be on any other libraries which depend on `bravado-core`.

Fixes https://github.com/Yelp/bravado/issues/307. I have also proposed a different fix for the same problem in https://github.com/Yelp/bravado/pull/320. I'm not sure if it's better to fix this in `bravado-core` or `bravado` (or both).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+- Sanitise resource names to ensure they are valid Python identifiers
+
 4.9.1 (2017-09-19)
 ------------------
 - Properly marshal a model even if it's not created from the same ``Spec`` instance - PR #194.
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+import re
 from collections import defaultdict
 
 from six import iteritems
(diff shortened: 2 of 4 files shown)


soumyakantiroychowdhury:
I also faced this issue and bypassed it using `getattr(client, 'My Tag').api_method().result()`. I am not in favor of your proposed change to rename 'My Tag` to `My_Tag` internally, if it breaks the existing code after upgrade.

Is it possible to do something like this:
`self.client.Resource('My Tag').api_method().result()`?

sjaensch:
Two comments:

1. The bravado SwaggerClient class already has a method ``_get_resource``, you can use it to fetch a resource with special characters. The underscore is just there to avoid name clashes, it's a public method.
2. This change is not backwards-compatible. It breaks existing code that uses`` _get_resource`` or the getattr trick mentioned by @soumyakantiroychowdhury to retrieve the resource.

I've tried a different approach in #243 that also sanitizes param names, let me know what you think.

PR #203, opened 2017-10-15
[CleanUp][Improvement] cached_property

macisamuele:
Use ``cached_property`` decorator instead of _private_ instance attributes to avoid re-execution of property methods.


bravado_core/operation.py
```diff
@@ -5,10 +5,29 @@
...
+    for regex, replacement in (
+            ('[^A-Za-z0-9_]', '_'),  # valid chars for method names
+            ('_+', '_'),  # collapse consecutive _'s
+            ('^_|_$', '')):  # trim leading/trailing _'s
+        operation_id = re.compile(regex).sub(replacement, operation_id)
+
+    # Handle crazy corner cases where someone explictily sets operation
+    # id a value that gets sanitized down to an empty string
+    if len(operation_id) == 0:
+        operation_id = _sanitize_operation_id(None, http_method, path_name)
```
sjaensch: Note that there is a (theoretical?) corner case where this leads to an infinite loop. If operation_id is already None we should throw an Exception or return a different result, but not recurse again.

tests/operation/operation_id_test.py
```diff
@@ -1,32 +1,22 @@
...
-from bravado_core.spec import Spec
+from bravado_core.operation import _sanitize_operation_id
 
 
 def test_returns_operation_id_from_operation_spec():
-    spec = Spec(spec_dict={})
-    operation_spec = {'operationId': 'getPetById'}
-    operation = Operation(spec, '/pet/{petId}', 'get', operation_spec)
-    assert 'getPetById' == operation.operation_id
+    assert 'getPetById' == _sanitize_operation_id('getPetById', 'get', '/pet/{petId}')
```
sjaensch: Why did you change these assertions?

bravado_core/util.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+
+class cached_property(object):
```
sjaensch: Can you add a test or two for this decorator? Especially to make sure it does not compute the value more than once.

tests/operation/operation_id_test.py
macisamuele: Changed the assertions because this tests are actually unittesting ``_sanitize_operation_id`` and in this way it is much more evident.

bravado_core/operation.py
macisamuele: ``operation_id`` is ``None`` if it is not specified in the swagger specs.
This method will replace ``operation_id`` with ``http_method + '_' + path_name`` (ie. ``test_returns_generated_operation_id_when_missing_from_operation_spec`` test) and the new operation ``id`` cannot be ``None`` or sanitised to an empty string.

bravado_core/util.py
macisamuele: added test 😄

bravado_core/operation.py
sjaensch: Well, if http_method is empty for any reason and path_name only contains '_' or invalid chars then this will result in an infinite loop I think. I agree that will probably never happen (outside of bugs), but let's make sure. :)

tests/operation/operation_id_test.py
sjaensch: But that's not what the test name says it tests (please change the name). We do want a test that verifies the property works correctly on the Spec class as well, right? It does dereferencing as well as calling `sanitize_operation_id`.

bravado_core/operation.py
macisamuele: This case is only possible in case of invalid specs.
We could be more conservative in this case by asserting that http_method is one of ``get``, ``put``, ``post``, ``delete``, ``options``, ``head``, ``patch`` (as specified by [pathItemObject](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject)).

sjaensch: Sounds good! I'm also happy with just throwing an exception in that case if it makes less work (simply because that is so much more obvious to debug than an infinite loop / hang). If we save the operation_id as passed in then we can check at this point if it was None originally - in that case there is no need for the recursive call, as it will not yield different results. We should just abort.

tests/util_test.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from bravado_core.util import cached_property
+
+
+def test_cached_property():
+    class Class(object):
+        calls = 0
```
sjaensch: This should be initialized in the \_\_init\_\_ method.

bravado_core/operation.py
macisamuele: Refactored a bit how it performs the sanitisation operations so we can prevent recursive calls :)

tests/operation/operation_id_test.py
macisamuele: Restored tests from Yelp/master

tests/util_test.py
macisamuele: 👍

Issue #209, opened 2017-11-01
we should support some notion of sensitive data in a spec


ajm188:
Say I have a spec that defines a field storing some PII. Ideally, if a request/response fails to validate, we would be able to stop the sensitive value from appearing in exception messages, logs, etc.

I propose adding some sort of "x-sensitive" property on a type schema.

ajm188:
Closed via #213

PR #213, opened 2017-11-02
Add a try/except to the validation

ajm188:
This will check against a new extension 'x-sensitive' and scrub the
value from the raised jsonschema.ValidationError if validation fails.

This addresses #209. All of the tests pass locally for me, and if we're satisfied with the approach, I'll add a test for the x-sensitive case.

### DIFF
--- a/bravado_core/validate.py
+++ b/bravado_core/validate.py
@@ -4,14 +4,35 @@ Delegate as much validation as possible out to jsonschema. This module serves
 as the single point of entry for validations should we need to further
 customize the behavior.
 """
+import sys
+from functools import wraps
+
 import jsonschema
 from six import itervalues
+from six import reraise
 
 from bravado_core.exception import SwaggerMappingError
 from bravado_core.exception import SwaggerSecurityValidationError
 from bravado_core.model import is_object
 from bravado_core.schema import SWAGGER_PRIMITIVES
 from bravado_core.swagger20_validator import get_validator_type
+
+
+def scrub_sensitive_value(func):
+    @wraps(func)
+    def scrubbed(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


bravado_core/validate.py
```diff
@@ -27,21 +28,27 @@ def validate_schema_object(swagger_spec, schema_object_spec, value):
...
-        raise SwaggerMappingError('Unknown type {0} for value {1}'.format(
-            obj_type, value))
+        else:
+            raise SwaggerMappingError('Unknown type {0} for value {1}'.format(
+                obj_type, value))
+    except jsonschema.ValidationError as e:
+        if e.schema.get('x-sensitive', False):
+            e.message = '*** ' + e.message[len(str(e.instance)):]
+            e.instance = '***'
+        raise
```
macisamuele: Please use six.reraise in order to preserve traceback.

bravado_core/validate.py
```diff
@@ -27,21 +28,27 @@ def validate_schema_object(swagger_spec, schema_object_spec, value):
     if not obj_type:
         return
 
-    if obj_type in SWAGGER_PRIMITIVES:
-        validate_primitive(swagger_spec, schema_object_spec, value)
+    try:
```
macisamuele: Instead of wrapping all in a big try/except statement could you wrap the method with a decorator (named like ``@hide_sensitive_info`` or similar)?

bravado_core/validate.py
ajm188: +1

bravado_core/validate.py
ajm188: Great idea!

PR #216, opened 2017-11-03
Fixes #214 - Added msgpack support

### LINKED ISSUE #214: Add msgpack support
msgpack would be a more efficient way to receive large JSON body responses, and could mean large performance gains for people who want to use bravado with a server that supports `application/msgpack` responses

Support for this content type should be added to bravado

AlstonLin:
### Test Plan
(1) `make test` with added unit tests
(2) Manually tested the client side by calling a mocked endpoint that responds with a application/msgpack content type and data, and verified the final response from Bravado was properly deserialized
(3) Manually tested the server side response validation by requesting an endpoint which responds with a application/msgpack content type that uses pyramid_swagger

### Related PRs
https://github.com/striglia/pyramid_swagger/pull/223
https://github.com/Yelp/bravado/pull/324

##### This should be released as a major version bump

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
+import umsgpack
 from six import iteritems
 
 from bravado_core.content_type import APP_JSON
+from bravado_core.content_type import APP_MSGPACK
 from bravado_core.exception import MatchingResponseNotFound
 from bravado_core.exception import SwaggerMappingError
 from bravado_core.unmarshal import unmarshal_schema_object
@@ -21,6 +23,7 @@ class IncomingResponse(object):
         'status_code',  # int - http status code
         'text',         # string - raw text of the body
         'headers',      # dict of http headers
+        'raw_bytes',    # the bytes of the body
     ]
 
     def __getattr__(self, name):
@@ -61,6 +64,7 @@ class OutgoingResponse(object):
         'content_type',  # str
         'text',          # str of body
         'headers',       # dict of headers
+        'raw_bytes',     # raw bytes of the response
     ]
 
     def __getattr__(self, name):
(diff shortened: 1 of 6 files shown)


bravado_core/response.py
```diff
@@ -103,15 +107,18 @@ def has_content(response_spec):
...
 
-    if content_type.startswith(APP_JSON):
+    if content_type.startswith(APP_JSON) or content_type.startswith(APP_MSGPACK):
         content_spec = deref(response_spec['schema'])
-        content_value = response.json()
-
+        if content_type.startswith(APP_JSON):
+            content_value = response.json()
+        else:
+            content_value = umsgpack.unpackb(response.raw_bytes)
```
macisamuele: I would suggest the usage of ``umsgpack.loads`` instead of ``unpackb`` to simplify understanding to who is used to JSON (library supports that too)

bravado_core/response.py
```diff
@@ -192,8 +199,11 @@ def validate_response_body(op, response_spec, response):
             "specification's content-types '{1}"
             .format(response.content_type, op.produces))
 
-    if response.content_type == APP_JSON:
-        response_value = response.json()
+    if response.content_type == APP_JSON or response.content_type == APP_MSGPACK:
+        if response.content_type == APP_JSON:
+            response_value = response.json()
+        else:
+            response_value = umsgpack.unpackb(response.raw_bytes)
```
macisamuele: I would suggest the usage of ``umsgpack.dumps`` instead of unpackb to simplify understanding to who is used to JSON (library supports that too)

setup.py
```diff
@@ -18,6 +18,7 @@
     "six",
     "swagger-spec-validator>=2.0.1",
     "pytz",
+    "u-msgpack-python>=2.4.1",
```
macisamuele: Could you make this requirement as loose as possible?
I guess that you're using features available on ``u-msgpack-python>=2``

tests/response/unmarshal_response_test.py
```diff
@@ -44,6 +46,20 @@ def test_json_content(empty_swagger_spec, response_spec):
...
 
+def test_msgpack_content(empty_swagger_spec, response_spec):
+    message = 'Monday'
+    response = Mock(
+        spec=IncomingResponse,
+        status_code=200,
+        headers={'content-type': APP_MSGPACK},
+        raw_bytes=umsgpack.packb(message))
+
+    with patch('bravado_core.response.get_response_spec') as m:
```
macisamuele: you can set the mock return value directly in patch definition
```python
with patch('bravado_core.response.get_response_spec', return_value=response_spec):
    ....
```

bravado_core/response.py
sjaensch: I have to say I disagree with that. We should be using the API the way it is intended to be used: i.e. it should feel correct and natural for people familiar with msgpack. and `unpackb` seems clearer to me: it conveys that we're dealing with binary data. Also, I don't think it will make a difference to people unfamiliar with msgpack. Did you have to look up what unpackb does @maci? :)

PR #218, opened 2017-11-04
Don't remove extraneous configurations, only warn about their presence

### LINKED ISSUE #217: 'also_return_response' should be added to CONFIG_DEFAULTS
This is currently breaking `make test` in bravado

```
E       KeyError: 'also_return_response'
```

This break is coming from https://github.com/Yelp/bravado-core/blob/master/bravado_core/spec.py#L146, which is deleting this key because it's not whitelisted in CONFIG_DEFAULTS

Alternatively, we could just not delete extra keys

macisamuele:
Fixes #217 .

The goal of this PR is to fix the current [``Yelp/bravado``](https://github.com/Yelp/bravado) [build failure](https://travis-ci.org/Yelp/bravado/builds/297225901).

``bravado`` is failing because it used to inject ``also_return_response`` configuration into ``bravado-core`` configs, which is not using that config.

Since ``bravado-core`` is not using this config, I'm not adding it into ``CONFIG_DEFAULTS``, but removing the logic that strips out extraneous configurations.

This PR exposes also ``_deref`` method in order to force dereferencing of objects even if ``internaly_dereference_ref`` in order to reduce ``test_deref_flattened_spec_not_recursive_specs`` flakyness.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -140,10 +140,9 @@ class Spec(object):
             are_config_changed = True
             for key in extraneous_keys:
                 warnings.warn(
-                    message='config {} is been removed because is not a recognized config key'.format(key),
+                    message='config {} is not a recognized config key'.format(key),
                     category=Warning,
                 )
-                del self.config[key]
 
         if self.config['internally_dereference_refs'] and not self.config['validate_swagger_spec']:
             are_config_changed = True
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -243,7 +243,7 @@ class Spec(object):
         else:
             return self.spec_dict
 
-    def deref(self, ref_dict):
+    def _force_deref(self, ref_dict):
         """Dereference ref_dict (if it is indeed a ref) and return what the
         ref points to.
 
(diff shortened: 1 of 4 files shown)


bravado_core/spec.py
```diff
@@ -244,17 +243,15 @@ def _internal_spec_dict(self):
         else:
             return self.spec_dict
 
-    def deref(self, ref_dict):
+    def _deref(self, ref_dict):
```
sjaensch: Hm, should we call this differently? E.g. `_force_deref`?

tests/spec/validate_config_test.py
```diff
@@ -21,22 +21,25 @@ def test_validate_config_succeed(minimal_swagger_dict, minimal_swagger_abspath,
...
 )
 @mock.patch('bravado_core.spec.warnings')
 def test_validate_config_fail(
-    mock_warnings, minimal_swagger_dict, minimal_swagger_abspath, config, expected_warnings_call,
+    mock_warnings, minimal_swagger_dict, minimal_swagger_abspath,
+    config, expected_different_config, expected_warnings_call,
 ):
     spec = Spec.from_dict(minimal_swagger_dict, origin_url=get_url(minimal_swagger_abspath), config=config)
-    assert spec.config != dict(CONFIG_DEFAULTS, **config)
+    assert bool(spec.config != dict(CONFIG_DEFAULTS, **config)) is expected_different_config
```
sjaensch: This needs the brackets, but not the bool cast (the brackets are needed due to Python's comparison operator chaining).

PR #224, opened 2017-11-14
allow to specify Content-Disposition filename

elmirjagudin:
I'm missing the ability to specify the original filename when uploading files with bravado client.

In this PR I have made it possible to specify filename by using a tuple as the argument to methods that POST files.

E.g. it possible to use:

client.foo.post_bar(fooFile=("my-filename", "file content"))

The old style is also supported:

client.foo.post_bar(fooFile="file content")

Not sure if using tuples is the best way to go, please advice.

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -284,7 +284,12 @@ def add_file(param, value, request):
                     param.op.consumes
             ))
 
-    file_tuple = (param.name, (param.name, value))
+    if isinstance(value, tuple):
+        filename, val = value
+    else:
+        filename, val = param.name, value
+
+    file_tuple = (param.name, (filename, val))
     request['files'].append(file_tuple)
 
 


macisamuele:
@elmirjagudin I've completely missed this PR ... I'm sorry about that 😞 

I'm not against this type of change but I would ask you to add some test coverage around your code change and to write an integration test for it. It's important for us to ensure that the added features are working as expected and to have automated ways to validate that.

Thanks a lot for contribuiting to the project.

elmirjagudin:
So you think it OK using tuples to specify file names.

I'll add tests, just did not want to spend time until I got an OK on the API.

elmirjagudin:
I realized I don't really have time finish up this PR. Hopefully I'll be back with new PR, with proper tests, at some later point.

PR #225, opened 2017-11-17
Return raw response

gstarnberger:
Adds a new config option "return_raw_response" that returns the raw decoded response (after json/msgpack decoding, skipping the call to unmarshal_schema_object).

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -115,6 +115,8 @@ def unmarshal_response(response, op):
             content_value = umsgpack.loads(response.raw_bytes)
         if op.swagger_spec.config['validate_responses']:
             validate_schema_object(op.swagger_spec, content_spec, content_value)
+        if op.swagger_spec.config['return_raw_response']:
+            return content_value
 
         return unmarshal_schema_object(
             op.swagger_spec, content_spec, content_value)
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -76,6 +76,9 @@ CONFIG_DEFAULTS = {
     # Completely dereference $refs to maximize marshaling and unmarshaling performances.
     # NOTE: this depends on validate_swagger_spec
     'internally_dereference_refs': False,
+
+    # Skip call to unmarshal_schema_object and return raw decoded response.
+    'return_raw_response': False,
 }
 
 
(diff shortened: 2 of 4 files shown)


bravado_core/response.py
```diff
@@ -115,6 +115,8 @@ def has_content(response_spec):
             content_value = umsgpack.loads(response.raw_bytes)
         if op.swagger_spec.config['validate_responses']:
             validate_schema_object(op.swagger_spec, content_spec, content_value)
+        if op.swagger_spec.config['return_raw_response']:
```
macisamuele: This config seems more related to "HTTP" handling and not swagger. I would suggest adding this into ``bravado`` implementation as [request option](https://github.com/Yelp/bravado/blob/master/bravado/client.py#L79) so you could enable this behavior on per request base.

gstarnberger:
I looked at request_options but I didn't find a way to implement the same functionality (without larger code changes). Essentially what I'd like to do is:

1. Add a new option that skips the call to _unmarshal_schema_object_ for latency-sensitive use-cases.

2. Still be able to enable options such as _validate_responses_ in dev/stage environments to verify that requests/responses match the spec.

All of this logic (response validation, response decoding, ...) is called in bravado_core.response.unmarshal_response so if I'd move the new code to a higher level I'd have to duplicate some of the unmarshal_response logic somewhere else (e.g., response validation, content-type detection, ...). For _return_raw_response_ I   followed the approach used for the _use_models_ config option which is similar as it also determines how the response is returned.

How would an implementation that uses request options look like? Would I have to duplicate the deserialization logic (json, msgpack) and the response validation logic somewhere else?

macisamuele:
@gstarnberger fair point ... During hackathon @sjaensch started to work a bit on async io bravado and we were discussing if it could make sense to move into bravado the logic for handling requests, so that could be already a sort of solution for you.

For your use case, I was thinking to add a request option (similar to ``also_return_response``) into bravado such that once it is enabled we're completely skipping the bravado-core interaction -> no validation, no unmarshaling and you still have access to the raw response (so bytes in case of msgpack).

Let me sync better with @sjaensch tomorrow morning, he could eventually have better ideas about your case 😄

PR #228, opened 2017-11-22
Fixes #227 - Use msgpack-python instead of umsgpack

### LINKED ISSUE #227: bravado_core should use the msgpack module instead of umsgpack
It seems that the umsgpack module, implemented in pure python, is an order of magnitude slower than the msgpack module, written with CPython bindings

AlstonLin:
Fixes #227 

### Testing Plan
- Updated unit tests and ran `make test`
- Tested server side response by using an API endpoint that renders `application/msgpack` responses and checked that the server side Swagger response validation worked properly
- Tested client side by using a `bravado` client which used a version of `bravado-core` with these changes to request an API endpoint that renders only `application/msgpack`. First, check that requesting without the `Accept: application/msgpack` results in a 500 response status. Then, check that requesting with the `Accept: application/msgpack` results in the desired response.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Changelog
 =========
+- Use msgpack-python instead of u-msgpack-python for performance improvements - Issue #227, PR #228.
 
 4.11.0 (2017-11-09)
 -------------------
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+4.12.0 (2017-11-22)
```
macisamuele: Please don't set release version and date since we have not released it yet.
I have no pypi access to release it

PR #232, opened 2017-12-06
Fixes #335 - Added support for validating compressed responses

### LINKED ISSUE #335: bravado-core does propagate x-nullable field in case of reference objects
While working on #331 I've noticed that bravado-core has some custom logic to ensure that the x-nullable field is propagated in case of a reference object.

According to the JSON schema definition a reference object is meant to fully replace the current object with the referenced one. This means that siblings attributes of a reference objects can not be considered (the definition does not strictly define this behaviour, so the decision is delegated to the implementer).

I'm mentioning this as ensuring the propagation of `x-nullable` results in
 * additional custom code to handle in the library
 * possible lack of feature parity between usage of `bravado_core.spec.Spec` with `internally_dereference_refs` set to `True` or `False`
 * possible confusion on the users as certain vendor extension will be propagated and others not (ie. a schema like `{'x-nullable': True, '$ref': '#/definitions/something', 'x-new': False}` will have `x-nullable` propagated and `x-new` ignored)

My suggestion would be to remove this manual handling and to bump swagger-spec-validator to 2.4.0+ which adds a warning in case a sibling is identified into a reference object (https://github.com/Yelp/swagger_spec_validator/pull/107)

⚠️ : this might break clients, so addressing this issue needs a major release

AlstonLin:
## Testing Plan
- Added unit tests and ran `make test`

### Manual Testing
Using bravado-core for Swagger validation and an API endpoint that produces the following:
- `Content-Type: application/msgpack` + `Content-Encoding: deflate` when `Accept: application/msgpack` and `Accept-Encoding: deflate` is specified
- `Content-Type: application/msgpack` when `Accept: application/msgpack` and is specified
- `Content-Type: application/json` + `Content-Encoding: deflate` when `Accept-Encoding: deflate` is specified
- `Content-Type: application/json` for any other request header combinations

The following was tested and verified, using a curl (to avoid any automatically added headers)
- `curl <URK> -i`
- `curl <URL> -H "Accept: application/msgpack" -i`
- `curl <URL> -H "Accept: application/msgpack" -H "Accept-Encoding: deflate" -i`
- `curl <URL> -H "Accept-Encoding: deflate" -i`

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -2,6 +2,7 @@
 import zlib
 
 import msgpack
+import simplejson as json
 from six import iteritems
 
 from bravado_core.content_encoding import DEFLATE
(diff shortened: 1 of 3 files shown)


bravado_core/response.py
```diff
@@ -203,7 +206,11 @@ def validate_response_body(op, response_spec, response):
         if response.content_type == APP_JSON:
             response_value = response.json()
         else:
-            response_value = msgpack.loads(response.raw_bytes, encoding='utf-8')
+            content_encoding = response.headers.get('content-encoding', '').lower()
+            raw_bytes = response.raw_bytes
+            if content_encoding == DEFLATE:
+                raw_bytes = zlib.decompress(raw_bytes)
```
macisamuele: Is there a reason for doing this only for msgpack content?
I would say that would be a great addition to do that for all the content types.

PR #234, opened 2018-01-08
Ensure bravado_core.spec.Spec object can be initialized with recursive spec and fully dereference

macisamuele:
The main goal of this PR is to ensure that Spec object can be initialized with recursive swagger specs and ``internally_dereference_refs`` set to ``True``.

The added test (``tests/spec/Spec/deref_flattened_spec_test.py::test_build_spec_object_with_recursive_specs_and_fully_dereference``) is going to fail if the changes added to ``bravado_core.spec.post_process_spec`` are reverted.

Additional changes provided by this CR are about:
 * logs: logs messages are built in ``logging`` library, this allows to avoid building strings that eventually are not used due to logging level
 * fix warnings during tests execution


bravado_core/resource.py
```diff
@@ -100,8 +100,7 @@ def __getattr__(self, item):
         """
         op = self.operations.get(item)
         if not op:
-            raise AttributeError(u"Resource '%s' has no operation '%s'" %
-                                 (self.name, item))
+            raise AttributeError(u"Resource '%s' has no operation '%s'", self.name, item)
```
sjaensch: This is not a call to a log function, so these two lines of code do different things.

macisamuele: 😢  ... too much sed :) thanks for checking it

PR #237, opened 2018-01-16
Percent encode params that will be put into the url

bschmaltz:
Before, we did not percent escape params being put into the url, so invalid urls would be created (most commonly with '%' or '?' characters).  When these invalid urls hit Apache, a 400 was returned, and bravado raised an exception when trying to parse the 400 (resulting in a 500 for the user).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+4.11.4 (2018-01-16)
+-------------------
+- Percent encode params that will be put into the url
+
 4.11.3 (2018-01-16)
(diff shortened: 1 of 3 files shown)


sjaensch:
Hey @bschmaltz, unfortunately I didn't have time to work on this today. I'd like to create a testcase and make sure that all HTTP clients handle this uniformly. I should have time to do it tomorrow. Thanks for submitting the bug report and the PR!

sjaensch:
Unfortunately the change to query param handling is not correct, I published #238 and have integration tests for bravado ready to publish once #238 is merged in.

PR #240, opened 2018-01-26
Fix 235 allow override formats

### LINKED ISSUE #235: Format double and overriding default formats
I have a use case that is using double numbers in the spec for monetary values.  Given that float is so imprecise, it would seem more valid to me to cast these to Decimal() from the decimal module.

btw, DEFAULT_FORMATS specifies both 'double' and 'float' format to the float()

Currently bravado-core does not allow you to override the DEFAULT_FORMATS, of course I can rename my spec format to decimal and get this to work for my implementation.  But I guess this brings up two questions:

1) Should you be able to override the DEFAULT_FORMATS in practice?
2) Should perhaps the default for double be to use the Decimal() class?

Thanks for sharing this awesome module.

bdowling:
This PR is to fix the need for overriding DEFAULT_FORMATS

I welcome feedback, particularly on the example provided in docs as I am not 100% clear on the validate function in particular, I haven't quite figured out when this is called.

Closes #235

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -314,9 +314,10 @@ class Spec(object):
         :param name: Name of the format to retrieve
         :rtype: :class:`bravado_core.formatters.SwaggerFormat`
         """
-        if name in formatter.DEFAULT_FORMATS:
-            return formatter.DEFAULT_FORMATS[name]
         format = self.user_defined_formats.get(name)
+        if format is None:
+            format = formatter.DEFAULT_FORMATS.get(name)
+
         if format is None:
             warnings.warn(
                 message='{0} format is not registered with bravado-core!'.format(name),
--- a/docs/source/formats.rst
+++ b/docs/source/formats.rst
@@ -168,3 +168,51 @@ CIDR objects back from the response.
     # Verify cidrs are CIDR objects and not strings
     for cidr in cidrs:
         assert type(cidr) == CIDR
+
+Overriding built-in formats is also possible with a user-defined format
+-----------------------------------------------------------------------
+By default format 'double' is internally converted to a float in python.  This runs
+the risk of being imprecise due to the way floats are handled.  If you would
+like to instead use the decimal.Decimal() type for more precision, you can override
+this built-in format.
+
+ `Floating Point Precision <https://docs.python.org/tutorial/floatingpoint.html>`_.
+
+You'd define the type:
+
+.. code-block:: python
+    from decimal import Decimal
+    import re
+
+    is_decimal = re.compile(r'^\d+(?:\.\d+)?$')
+    def validate_decimaltype(x):
+      """Validate input is a str in valid decimal format"""
+      if not (isinstance(x, str) and is_decimal.match(x)):
+          raise bravado_core.exception.SwaggerValidationError()
+
+    mydouble = SwaggerFormat(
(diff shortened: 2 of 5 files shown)


bravado_core/spec.py
```diff
@@ -314,9 +314,9 @@ def get_format(self, name):
         :param name: Name of the format to retrieve
         :rtype: :class:`bravado_core.formatters.SwaggerFormat`
         """
-        if name in formatter.DEFAULT_FORMATS:
-            return formatter.DEFAULT_FORMATS[name]
-        format = self.user_defined_formats.get(name)
+        format = self.user_defined_formats.get(name,
```
macisamuele: To make it easier to read (and not use too long lines) I would suggest to first fetch the formatter from the user defined formmatters and if nothing is extracted trying to get it from the default formats
```python
format = self.user_defined_formats.get(name)
if format is None:
    format = formatter.DEFAULT_FORMATS.get(name)
```

docs/source/formats.rst
```diff
@@ -168,3 +168,43 @@ CIDR objects back from the response.
     # Verify cidrs are CIDR objects and not strings
     for cidr in cidrs:
         assert type(cidr) == CIDR
+
+Overriding built-in fomrats is also possible with a user-defined format
+-----------------------------------------------------------------------
+By default 'double' is internally converted to a float in python.  This runs
```
macisamuele: I would mention "double format"

docs/source/formats.rst
```diff
@@ -168,3 +168,43 @@ CIDR objects back from the response.
...
+ `Floating Point Precision <https://docs.python.org/tutorial/floatingpoint.html>`_.
+
+You'd define the type:
+
+.. code-block:: python
+    from decimal import Decimal
+
+    def mydouble_validate(x):
+	"""Validate input is in fact a decimal type"""
+	if isinstance(x,Decimal):
```
macisamuele: The validate method gets the "json" object representing your attribute.
So in your case, it will be something like
```yaml
...
    type: number
    format: double
```
It means that ``type(x)`` will be a ``float``

tests/formatter/to_wire_test.py
```diff
@@ -125,3 +126,30 @@ def test_ref(minimal_swagger_dict):
     result = to_wire(swagger_spec, int_ref_spec, 999)
     assert 999 == result
     assert isinstance(result, int)
+
+
+def test_override(minimal_swagger_dict):
```
macisamuele: I wonder if we can simplify this test

bravado_core/spec.py
bdowling: I realized the schema is actually be string(double), so json should leave it alone until it gets marshalled.

docs/source/formats.rst
bdowling: I realized the schema is actually be string(double), so json should leave it alone until it gets marshalled.

tests/formatter/to_wire_test.py
bdowling: Git is not seeing this as changed from where you put tagged the comment, but I think the new test is a simplified MVT.

bdowling:
@macisamuele - I re-based squashed and tossed up this PR, let me know if there is anything else.

PR #242, opened 2018-01-31
Change model repr() to return additionalProperties

bdowling:
I had a case where I was using `additionalProperties: true` (actually that's the default I think) and I noticed that the repr() on the response did not include them in the object.  It includes them in the dict when `use_models: false` or _as_dict() is called.

Should it include them in the repr(Model) ?


sjaensch:
Makes sense to me, thanks for your contribution!

Issue #215, opened 2017-11-03
Ensure param names and tags are valid python identifiers


bplotnick:
We sanitize [operation_ids](https://github.com/Yelp/bravado-core/blob/master/bravado_core/operation.py#L13), but we should also sanitize param names and tags to ensure that they are valid python parameters.

There is a workaround as mentioned in https://github.com/Yelp/bravado/issues/293, but we should just fix it here.

This might have some backwards compatibility considerations, but I don't think anything serious since resources and params that are currently not valid python identifiers are not even accessible via the normal means currently (i.e. via dot or method params)

PR #243, opened 2018-02-05
Sanitize param and resource names

sjaensch:
Currently, bravado-core only makes sure the operationId is a valid Python identifier. Let's do the same for resources and params. My primary concern was backwards compatibility; I've ran bravado tests with this bravado-core branch and they all passed, including a few new integration tests that test the new behavior. I've used this branch with the bravado-asyncio test suite as well, all tests pass.

The second concern was discoverability. The names shown when using dir(client) or dir(resource) in bravado need to be the sanitized names. That's why the real name is only added as an alias.

Closes #200, #215.

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -1,31 +1,26 @@
 # -*- coding: utf-8 -*-
 import logging
-import re
+
+from six import iteritems
 
 from bravado_core.exception import SwaggerSchemaError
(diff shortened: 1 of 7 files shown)


macisamuele:
@sjaensch I really like your approach ... this will be able to provide sanitized names without disrupting users that are using ``getattr`` workaround.

Please fix tests and feel free to merge it 😄

PR #246, opened 2018-02-07
Collect models will not create models for not models objects

macisamuele:
Current model discovery generates models also for _not models_ object specs.
This PR aims to solve this issue.

Assuming that the following is a snippet of the swagger specs
```json
{
    "definitions": {
        "list": {
            "type": "array",
            "items": {"type": "string"},
            "x-model": "list"
        }
    }
    ...
}
```
Before this PR this generates a ``list`` model (``assert Swagger.from_url(...).definitions == {'list': mock.ANY}``)
After this PR ``assert Swagger.from_url(...).definitions == {}``

### DIFF
--- a/tests/model/collect_models_test.py
+++ b/tests/model/collect_models_test.py
@@ -30,3 +30,23 @@ def test_simple(minimal_swagger_dict, pet_model_spec):
         models=models,
         swagger_spec=swagger_spec)
     assert 'Pet' in models
+
+
+def test_not_model(minimal_swagger_dict, pet_model_spec):
+    minimal_swagger_dict['definitions']['Pet'] = pet_model_spec
+    minimal_swagger_dict['definitions']['Pets'] = {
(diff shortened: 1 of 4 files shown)


tests/model/collect_models_test.py
```diff
@@ -30,3 +30,23 @@ def test_simple(minimal_swagger_dict, pet_model_spec):
         models=models,
         swagger_spec=swagger_spec)
     assert 'Pet' in models
+
+
+def test_not_model(minimal_swagger_dict, pet_model_spec):
+    minimal_swagger_dict['definitions']['Pet'] = pet_model_spec
```
sjaensch: Maybe add a docstring explaining why we this is not a model, although it has the x-model marker?

macisamuele: Added docstring and updated documentation

PR #247, opened 2018-02-09
deref is overridden after building Spec

macisamuele:
In order to optimize dereferencing process, we're dynamically defining ``deref`` method according to ``internally_dereference_refs`` value.
This implies that during spec post-processing the no-op deref will be used if ``internally_dereference_refs`` is enabled.

This CR fixes this issue by:
 * forcing ``deref`` to be an alias of ``_force_deref``
 * overriding ``deref`` (to be a _no-op_) only after spec post-processing (model discovery) runs over the specs

An additional change provided by this CR is related to how models are saved in case of ``internally_dereference_refs``.

Before this change, we were running ``descend`` (on ``post_process_spec``) twice to guarantee that the models get's overridden; but I noticed that this was not happening for models that are not directly referenced (ie. models defined for polymorphism)

After this change, we build a new temporary ``Spec`` object which guarantees that model discovery is performed on fully dereferenced objects.

I don't expect this change to be a breaking change and to add additional time penalties in building ``Spec`` objects.

NOTE: I've also updated ``tests/profiling/unmarshal_response_profiler_test.py`` so that executions gets faster if benchmark tests are disabled.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -125,13 +125,6 @@ class Spec(object):
 
         self._validate_config()
 
-        if self.config['internally_dereference_refs']:
-            # If internally_dereference_refs is enabled we do NOT need to resolve references anymore
-            # it's useless to evaluate is_ref every time
-            self.deref = lambda ref_dict: ref_dict
-        else:
-            self.deref = self._force_deref
-
     def _validate_config(self):
         """
         Validates the correctness of the configurations injected and makes sure that:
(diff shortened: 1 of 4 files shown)


bravado_core/spec.py
```diff
@@ -222,21 +215,40 @@ def _validate_spec(self):
...
+                        swagger_spec=self,
+                    ),
+                    functools.partial(
+                        collect_models,
+                        models=swagger_spec.definitions,
+                        swagger_spec=self,
+                    ),
+                ],
+            )
+            return swagger_spec.definitions
```
sjaensch: This seems like a weird API to me. Nothing in the name suggests that the function returns a part of the processed spec. I'd either completely remove it or return the full spec.

bravado_core/spec.py
```diff
@@ -222,21 +215,40 @@ def _validate_spec(self):
...
+        # discovery is performed
+        run_post_processing(self)
+
+        if self.config['internally_dereference_refs']:
+            deref_flattened_spec = self.deref_flattened_spec
+            tmp_spec = Spec(deref_flattened_spec, self.origin_url, self.http_client, self.config)
+
+            # Rebuild definitions using dereferences specs as base
+            # this ensures that the generated models have no references
+            self.definitions = run_post_processing(tmp_spec)
```
sjaensch: I'm assuming you did it because you need the definitions here? I'd just do it in two lines.

bravado_core/spec.py
macisamuele: I can easily remove the return statement from here ... it was a convenience for line 248 😄 
I'm splitting it in 2 lines

macisamuele:
Fair point about adding a test to prevent regression.
I've added ``test_ensure_model_spec_contains_reference_if_fully_dereference_is_not_enabled`` and ``test_ensure_model_spec_does_not_contain_references_if_fully_dereference_is_enabled`` to guarantee that in case of regression tests will fail

PR #248, opened 2018-02-09
Workaourd to fix internal build

macisamuele:
Our internal build is failing on master 😢 

I tried to inspect the error and seems that setuptools is complaining when trying to install ``webcolors==1.8``.
I'm proposing this in order to fix our internal build.

NOTE: ``webcolor`` is not a direct dependency (it is added from ``jsonschema[format]``, so I'm not to concerned in not using the latest version as I assume it is already tested there.


sjaensch:
We've fixed it internally.

PR #249, opened 2018-02-09
Improve model name determination by using title information

macisamuele:
This CR improves the model name determination strategy used during model discovery.

The original approach was:
 * use ``key`` for models in ``definition``
 * use ``x-model`` for models around the specs

The goal of this PR tries to unify this in a common approach.
The model name is determined as:
 1. use ``x-model``
 2. use ``title``
 3. if the model is in ``#/definitions`` use the ``key``

WARNING: this does CR not try to improve detection of duplicated models, but it should simplify future work on that direction

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -125,7 +125,10 @@ def bless_models(container, key, path, visited_models, swagger_spec):
 
     if (
         not is_dict_like(model_spec) or
-        not is_object(swagger_spec, model_spec, ignore_config=True) or
+        not is_object(swagger_spec, model_spec, no_default_type=True) or
+        # NOTE: determine_object_type uses a simple heuristic to determine if a model_spec has a SCHEMA type
+        # for this reason is important that model_spec is recognized as model in the most accurate way
+        # so we should not rely on default typing of a schema
         determine_object_type(model_spec) != ObjectType.SCHEMA or
         deref(model_spec.get(MODEL_MARKER)) is not None
     ):
@@ -547,19 +550,19 @@ def is_model(swagger_spec, schema_object_spec):
     return deref(schema_object_spec.get(MODEL_MARKER)) is not None
 
 
-def is_object(swagger_spec, object_spec, ignore_config=False):
+def is_object(swagger_spec, object_spec, no_default_type=False):
     """
     A schema definition is of type object if its type is object or if it uses
     model composition (i.e. it has an allOf property).
     :param swagger_spec: :class:`bravado_core.spec.Spec`
     :param object_spec: specification for a swagger object
     :type object_spec: dict
-    :param ignore_config: ignore bravado-core 'default_type_to_object' configuration
-    :type ignore_config: bool
+    :param no_default_type: ignore bravado-core 'default_type_to_object' configuration
+    :type no_default_type: bool
     :return: True if the spec describes an object, False otherwise.
     """
     deref = swagger_spec.deref
-    default_type = 'object' if not ignore_config and swagger_spec.config['default_type_to_object'] else None
+    default_type = 'object' if not no_default_type and swagger_spec.config['default_type_to_object'] else None
     return deref(object_spec.get('type', default_type)) == 'object' or 'allOf' in object_spec
 
 
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -7,8 +7,11 @@ import six
 from six import iteritems
 
 from bravado_core.schema import collapsed_properties
+from bravado_core.schema import is_dict_like
 from bravado_core.schema import is_list_like
 from bravado_core.schema import SWAGGER_PRIMITIVES
+from bravado_core.util import determine_object_type
+from bravado_core.util import ObjectType
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


bravado_core/model.py
```diff
@@ -63,12 +72,11 @@ def collect_models(container, key, path, models, swagger_spec):
     :param models: created model types are placed here
     :type swagger_spec: :class:`bravado_core.spec.Spec`
     """
-    deref = swagger_spec.deref
-    if key == MODEL_MARKER and is_object(swagger_spec, container):
+    if (key == MODEL_MARKER or key == 'title') and is_object(swagger_spec, container):
```
sjaensch: While you didn't write this code, it is quite confusing to me what `key` really is. From looking at how `tag_models`and `collect_models` is called, it should be either the dict key with which the model instance is stored, or a list index. How can it ever be MODEL_MARKER or 'title'? And conceptually, I don't think it should be both - it should be either one or the other.

macisamuele: I've added ``key=='title'`` in order to recognize inline models (see example below), but I agree that this increases understanding complexity.
I've also noticed that bravado-core relies on ``MODEL_MARKER`` to determine if a model type can be used (check ``bravado_core.model.is_model``, so the added complexity will not produce the expected behavior during response unmarshaling.

In order to keep the code simpler to read I'm removing ``key=='title'`` from the callback (so inline schemas are not tagged as models -> status-quo), adding docstrings to better expose the  expected callback _environment_. I'll add an extra callback (called ``bless_models`` that will take care of tagging inline models).

Additionally, we don't need to check for ``key=='title'`` because ``tag_models`` will tag models that have no ``x-model`` defined.

macisamuele:
I've added documentation for the callbacks methods and added a new callback method called ``bless_models``.
The new callback is in charge of tagging models discovered during spec traversing.
The difference between ``tag_models`` and ``bless_models`` is that:
 * ``tag_models`` tags models present on ``...#/definitions`` defaulting the model name (if not present) to the definition ``key``
 * ``bless_models`` tags all the models identified during spec traversing (inline models). If an inline model is present but no model name is assigned (``x-model`` and ``title`` are missing) no tag will be added

bravado_core/model.py
```diff
@@ -457,17 +547,19 @@ def is_model(swagger_spec, schema_object_spec):
...
+def is_object(swagger_spec, object_spec, ignore_config=False):
     """
     A schema definition is of type object if its type is object or if it uses
     model composition (i.e. it has an allOf property).
     :param swagger_spec: :class:`bravado_core.spec.Spec`
-    :param schema_object_spec: specification for a swagger object
-    :type schema_object_spec: dict
+    :param object_spec: specification for a swagger object
+    :type object_spec: dict
+    :param ignore_config: ignore bravado-core 'default_type_to_object' configuration
```
sjaensch: You're not ignoring the config, you're behaving as if the setting were False. Maybe that would be a better way to name and describe this parameter. ``no_default_type``? ``disable_default_type``?

bravado_core/model.py
```diff
@@ -42,33 +92,73 @@ def tag_models(container, key, path, visited_models, swagger_spec):
...
+    """
+    if not is_dict_like(container):
+        return
+
+    deref = swagger_spec.deref
+    model_spec = deref(container.get(key))
+
+    if (
+        not is_dict_like(model_spec) or
+        not is_object(swagger_spec, model_spec, ignore_config=True) or
```
sjaensch: It is very much unclear as to why you'd need to ignore the config (i.e. not use a default `object` type) in this one specific instance.

bravado_core/swagger20_validator.py
```diff
@@ -150,14 +151,14 @@ def discriminator_validator(swagger_spec, validator, discriminator_attribute, in
             message='\'{}\' is not a recognized schema'.format(discriminator_value)
         )
 
-    if discriminator_value == schema['x-model']:
+    if discriminator_value == schema[MODEL_MARKER]:
```
sjaensch: I think using `'x-model'` instead of the constant is a feature: Changing MODEL_MARKER should break all sorts of tests. Using these unchangeable constants in test code reduces the value of the test. I do not want to make sure that the code I test uses the constant. I want to make sure it uses `'x-model'`.

sjaensch: Sorry, commented on the wrong piece of code, but you get the idea. :)

bravado_core/model.py
macisamuele: If ``default_type_to_object`` is set to true then ``is_object(swagger_spec, model_spec)`` returns ``True`` for a big variety of objects (ie. ``#/definitions`` will always be seen an object).

Adding the parameter allows us to not duplicate the check on ``type=="object" or "allOf" in model_spec``

I'll add a comment to better document this

bravado_core/model.py
macisamuele: I'll rename this in ``no_default_type ``

bravado_core/swagger20_validator.py
macisamuele: ack ... I'll _revert_ this change from tests

PR #250, opened 2018-02-12
Small tests refactory to ensure high testing coverage

macisamuele:
This PR removes not used fixtures and makes sure that all the tests get executed.
I've noticed this while checking coverage of the test related code.

ℹ️ : I've moved ``test_security_object_and_definition_constants`` to ensure its execution (it was in conftest and seemed that pytest was not running the test)

### DIFF
--- a/tests/operation/security_object_test.py
+++ b/tests/operation/security_object_test.py
@@ -9,6 +9,13 @@ from bravado_core.request import IncomingRequest
 from bravado_core.request import unmarshal_request
 from bravado_core.resource import build_resources
 from bravado_core.spec import Spec
+from tests.operation.conftest import SECURITY_DEFINITIONS
+from tests.operation.conftest import SECURITY_OBJECTS
+
+
+def test_security_object_and_definition_constants():
+    """Test that ensures that constant in tests/operation/conftest.py are reasonable"""
+    assert SECURITY_OBJECTS.keys() == SECURITY_DEFINITIONS.keys()
 
 
 def test_op_with_security_in_op_without_security_defs(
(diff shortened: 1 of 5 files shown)


tests/operation/security_object_test.py
```diff
@@ -9,6 +9,13 @@
 from bravado_core.request import unmarshal_request
 from bravado_core.resource import build_resources
 from bravado_core.spec import Spec
+from tests.operation.conftest import SECURITY_DEFINITIONS
+from tests.operation.conftest import SECURITY_OBJECTS
```
sjaensch: Shouldn't these be pytest fixtures?

tests/operation/security_object_test.py
```diff
@@ -9,6 +9,13 @@
...
+
+
+def test_security_object_and_definition_constants():
+    """Test that ensures that constant in tests/operation/conftest.py are reasonable"""
+    assert SECURITY_OBJECTS.keys() == SECURITY_DEFINITIONS.keys()
```
sjaensch: What we do in other projects is define such constants in testing/fixtures.py and then have tests/testing/fixtures_test.py where we do these kinds of tests. That said it seems a bit overkill in this case, just leave it as-is.

tests/operation/security_object_test.py
macisamuele: The two constant dicts are used only in ``conftest.py`` to generate fixtures ...
Defining those as fixtures will make the other fixtures definition even _more complex_.

I can eventually suggest to remove ``test_security_object_and_definition_constants`` since it tests that tests data are complete ... maybe too _meta_ 😄

PR #253, opened 2018-02-19
Duplicate models should not raise if not use models

macisamuele:
The current implementation of ``bravado-core`` raises ``ValueError`` exception if ``tag_models`` or ``bless_models`` identifies a duplicated model.

This behavior is correct is ``use_models`` config is set to ``True``, but it is erroneous if no models are required to the ``Spec`` object.

This PR tries to solve this issue.

### DIFF
--- a/tests/model/tag_models_test.py
+++ b/tests/model/tag_models_test.py
@@ -81,6 +81,8 @@ def test_duplicate_model(mock_log, minimal_swagger_dict, pet_model_spec, use_mod
 
     duplicate_message = 'Duplicate "Pet" model found at path [\'definitions\', \'Pet\']. ' \
                         'Original "Pet" model at path [\'definitions\', \'Pet\']'
+
+    raised_exception = None
     try:
         tag_models(
             minimal_swagger_dict['definitions'],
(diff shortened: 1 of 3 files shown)


tests/model/tag_models_test.py
```diff
@@ -71,17 +73,27 @@ def test_path_too_short(minimal_swagger_dict, pet_model_spec):
...
             visited_models={'Pet': ['definitions', 'Pet']},
-            swagger_spec=swagger_spec)
-    assert 'Duplicate' in str(excinfo.value)
+            swagger_spec=swagger_spec,
+        )
+    except ValueError as e:
+        assert str(e) == duplicate_message
+        assert not mock_log.warning.called
+    else:
+        mock_log.warning.assert_called_once_with(duplicate_message)
```
sjaensch: Well, in an effort to not duplicate code you've written a test that doesn't tell us when (if ever) the exception is raised, and when we do logging. I think it's ok in this case, just something to be aware of.

macisamuele: Fixed it

PR #255, opened 2018-02-20
Generate model names

macisamuele:
This PR adds model name generation to bravado-core model discovery system.

The current master is not generating model names (tagging the model with ``x-model``) if it is not possible to generate a model name from the model spec (so using ``x-model``, ``title`` and eventually definition ``key``).
The goal of this PR is to change this by generating model names for models that have not enough information to have a model name *pre-assigned* to them.

**How this could be achieved**
To generate model names we need to:
 * modify ``bless_models`` to generate a model name in case ``_get_model_name`` returns no name
 * in case of models in ``definitions`` we should prefer ``key`` respect the generated model name (to be consistent with the documentation and reduce breakage) -> generated model names are tracked as blessed and un-registered if needed (``_unregister_visited_model``)
 * spec post processing should perform two spec traversal rounds: 1 for tagging the models and 1 for generating the model types. This is needed to guarantee precedence of tagged models over blessed models. This could be seen as a performance hit during spec build.
  ⚠️  I will probably reconsider this as PR #254 guarantees that specs are traversed following a pre-defined order that gives priority to ``definitions`` over ``paths``

NOTES:
 * The generated model name has to be as unique. Using path should fix this but names clashing are still possible. PR #253 ensures that duplicated model names are not throwing exceptions if models are disabled. This gives us time to fix the specs (if we have the possibility to do that) or still be able to use ``bravado-core`` by giving up on models.

I'm not super sure yet of this changes, this PR has been published to gather comments 😄 

Before pushing this PR I do need to
 * update changelog
 * update models documentation page

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -12,13 +12,16 @@ from bravado_core.schema import is_list_like
 from bravado_core.schema import SWAGGER_PRIMITIVES
 from bravado_core.util import determine_object_type
 from bravado_core.util import ObjectType
+from bravado_core.util import sanitize_name
 from bravado_core.util import strip_xscope
+
 
 log = logging.getLogger(__name__)
 
 # Models in #/definitions are tagged with this key so that they can be
 # differentiated from 'object' types.
 MODEL_MARKER = 'x-model'
+MODEL_MARKER_BLESSED = 'x-model-blessed'
 
 
 def _get_model_name(model_dict):
@@ -27,6 +30,10 @@ def _get_model_name(model_dict):
     if not model_name:
         model_name = model_dict.get('title')
     return model_name
+
+
+def _generate_model_name(path):
+    return sanitize_name('_'.join(path))
 
 
 def _raise_or_warn_duplicated_model(swagger_spec, message):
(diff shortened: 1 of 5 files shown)


bravado_core/model.py
```diff
@@ -12,13 +12,16 @@
...
+from bravado_core.util import sanitize_name
 from bravado_core.util import strip_xscope
 
+
 log = logging.getLogger(__name__)
 
 # Models in #/definitions are tagged with this key so that they can be
 # differentiated from 'object' types.
 MODEL_MARKER = 'x-model'
+MODEL_MARKER_BLESSED = 'x-model-blessed'
```
sjaensch: I'm not a fan of yet another property that we add to the spec itself; in addition, what "blessed" means is entirely unclear and doesn't explain what the tag does.

macisamuele:
As agreed with @sjaensch before eventually thinking to apply this PR we should ensure that models present on definitions of the referred files are kept into consideration (as stated by the documentation).
I'm closing this PR for now, if later we figure out that model name generation is needed we can work on it.

Issue #258, opened 2018-03-02
bravado-core does not support multiple parameters properly


nickdirienzo:
We're experiencing an issue where `cast_request_param` is failing with a `TypeError` in the following scenario. For the record, we use pyramid_swagger and by virtue of that, bravado-core.

If you have a spec like:
```
"/objects":
    get:
      operationId: list_objects
      tags:
        - Events
      parameters:
        - in: query
          name: parent_id
          required: true
          description: The ID of the Parent you would like to list all Objects for
          type: integer
          format: int64
```

Then make a request with:
```
GET /resource?parent_id=123&parent_id=456
```

bravado-core will raise a `TypeError` and therefore a 500 because we're trying to cast `['123', '456']` to via `int()`.

I may be misunderstanding how [parameter descriptions](https://swagger.io/docs/specification/2-0/describing-parameters/) work in Swagger 2.0, but from my perspective, this seems like a problem in the library.

A simple fix to not return 500s, would be to also catch `TypeError` and then allow jsonschema to handle the validation (as suggested by the comment [here](https://github.com/Yelp/bravado-core/blob/54d099e3fa1df0fe027908e96dd3718b11b689bf/bravado_core/param.py#L257)), which would then would return a 400 instead. This seems more proper. I'll put up a PR to address the issue, but happy to discuss further in case I'm mistaken.

Looking forward to hearing from you @sjaensch!

nickdirienzo:
Hmm, thinking about this a bit more, I'm not sure if this is an issue with bravado-core or pyramid_swagger given that the docstring of `cast_request_param` is:
```
Try to cast a request param (e.g. query arg, POST data) from a string to
    its specified type in the schema. This allows validating non-string params.

:param param_type: name of the type to be casted to
    :type  param_type: string
    :param param_name: param name
    :type  param_name: string
    :param param_value: param value
    :type  param_value: string
```

So we shouldn't ever be sending a list into this function in the first place. I think I need to dig a bit more here.

nickdirienzo:
Ah okay. This seems like that's the root cause now. As long as the `cast_param` function doesn't raise an exception we will eventually hit the collection format line [here](https://github.com/Yelp/bravado-core/blob/54d099e3fa1df0fe027908e96dd3718b11b689bf/bravado_core/param.py#L194), which handles collections properly. Unfortunately, we're raising a `TypeError` and bombing out before we should.

[...]

PR #259, opened 2018-03-02
If type casting fails on a ``TypeError`` allow JSON Schema to handle the error

nickdirienzo:
In #258, it was discovered that multiple query parameters of a primitive type cause a `TypeError` causing an application server to 500 instead of 400. This change catches the `TypeError`, allowing JSON Schema to handle the exception and return a 400 instead.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


nickdirienzo:
Thanks @sjaensch for the review, merge, and publish! Really appreciate the fast turnaround. 👍

PR #261, opened 2018-04-03
Don't require jsonschema[formats]

ralsina:
If you require that, jsonschema requires rfc3987 and strict_rfc3339 which are licensed under the GPL, rendering bravado-core and bravado effectively GPL-licensed.

Which is **bad**.


sjaensch:
@ralsina while you're raising a good point we in fact need jsonschema[format] to correctly validate the date-time format. Any suggestions on how to avoid the dependency?

ralsina:
If you know how to validate it other than using GPL code,  adding a checker to jsonschema looks pretty easy:

https://github.com/Julian/jsonschema/blob/118e6a2412afcb64544c607425fec2ef346880bf/jsonschema/_format.py

It could be done by proposing a PR for jsonschema or on the bravado-core side.

Or you could add an optional [formats] to your setup.py too so users can choose instead of having to fork it to avoid this issue.

In the meantime your library is, for all practical purposes, GPL.

ralsina:
@sjaensch there you have a change that uses a MIT-licensed library to check date-time values. Change is minimal but it may not be done in the right place, did not want to dig too much.

Hope that helps.

bravado_core/swagger20_validator.py
```diff
@@ -32,6 +34,11 @@ def format_validator(swagger_spec, validator, format, instance, schema):
     if (is_param_spec(swagger_spec, schema) or
             is_prop_nullable(swagger_spec, schema)) and instance is None:
         return
+    elif format == 'date-time':
```
sjaensch: I don't think it's necessary to do this here; rather, we should modify bravado_core/formatter.py and specify the validation format there for `date-time`.
I'm not sure if removing jsonschema[format] means we also need to manually validate the `date` format...?

ralsina: "I'm not sure if removing jsonschema[format] means we also need to manually validate the date format...?"

Probably not, I think. Or at least, no test for it failed.

ralsina:
@sjaensch you're welcome. I can't spend any more time in it tho, so feel free to take over if the change is in the wrong place (my company moved to an internal fork until this is fixed)

bravado_core/swagger20_validator.py
macisamuele: Checking jsonschema (extra dependencies](https://github.com/Julian/jsonschema/blob/master/setup.py#L46) I see that ``format`` adds
 * rfc3987 : for uri format (https://github.com/Julian/jsonschema/blob/master/jsonschema/_format.py#L256)
 * strict-rfc3339  : for date-time format (https://github.com/Julian/jsonschema/blob/master/jsonschema/_format.py#L223)
 * webcolors : for color format (https://github.com/Julian/jsonschema/blob/master/jsonschema/_format.py#L256)

If we're removing ``format`` extra dependency from ``jsonschema`` we should replace the validators with more license friendly ones

ralsina: * webcolors is BSD so it's probably ok.
* strict-rfc3339 is what I replaced.
* rfc3987 is probably not worth checking? In any case, there is no validator under a non-copyleft license I can see. Maybe you can just validate using python's stdlib url parser.

ralsina:
Over a year. Closing this, life is too short. I am not even working at the company that used this anymore.

PR #263, opened 2018-04-24
Fix spec flattening in case of relative references

macisamuele:
While working on internal specs I've spotted two issues related to flattening and dereferencing specs with relative links on different directories.

Everything works fine if the references are not on different files, but as soon as we reference a file on a different directory all the links on the new files are relative to the new path.

The issue is caused because:
 * reference resolving does not take in consideration the resolution scope (changes on `bravado_core/spec_flattening.py`)
 * flattening specs requires resources to be available; as resources are not built at the time of `flattened_spec` call, enforcing resource building as an extra step of `run_post_processing` fixes the issue

**Note**:
while doing those changes I've also noted that `_internal_spec_dict` was returning `deref_flattened_spec` if `internally_dereference_refs` config is enabled.
Sadly the evaluation of `deref_flattened_spec` requires resource building, which requires `_internal_spec_dict`, so an ugly cycle of badly handled dependencies.
In order to decouple these dependencies and remove the minimal overhead related to property evaluation I've converted `_internal_spec_dict` property to Spec attribute, which is overridden in case `internally_dereference_refs` config is set

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -124,6 +124,7 @@ class Spec(object):
             handlers=build_http_handlers(http_client),
         )
 
+        self._internal_spec_dict = spec_dict
         self._validate_config()
 
     def _validate_config(self):
@@ -259,18 +260,12 @@ class Spec(object):
 
             # Avoid to evaluate is_ref every time, no references are possible at this time
             self.deref = lambda ref_dict: ref_dict
+            self._internal_spec_dict = self.deref_flattened_spec
 
         for format in self.config['formats']:
             self.register_format(format)
 
         self.api_url = build_api_serving_url(self.spec_dict, self.origin_url)
-
-    @cached_property
-    def _internal_spec_dict(self):
-        if self.config['internally_dereference_refs']:
-            return self.deref_flattened_spec
-        else:
-            return self.spec_dict
 
     def _force_deref(self, ref_dict):
         """Dereference ref_dict (if it is indeed a ref) and return what the
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -368,12 +368,8 @@ class Spec(object):
 
         return strip_xscope(
             spec_dict=flattened_spec(
-                spec_dict=self.spec_dict,
-                spec_resolver=self.resolver,
-                spec_url=self.origin_url,
-                http_handlers=build_http_handlers(self.http_client),
-                spec_definitions=self.definitions,
                 swagger_spec=self,
+                http_handlers=build_http_handlers(self.http_client),
             ),
         )
 
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -124,6 +124,8 @@ class Spec(object):
             handlers=build_http_handlers(http_client),
         )
 
+        # spec dict used to build resources, in case internally_dereference_refs config is enabled
+        # it will be overridden by the dereferenced specs (by build method). More context in PR#263
         self._internal_spec_dict = spec_dict
         self._validate_config()
 
@@ -246,6 +248,8 @@ class Spec(object):
         # deref_flattened_spec depends on flattened_spec which assumes that model
         # discovery is performed
         run_post_processing(self)
+        # Flattening the specs requires resources to be available.
+        # Let's build them before self.deref_flattened_spec is called
         self.resources = build_resources(self)
 
         if self.config['internally_dereference_refs']:
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -245,6 +245,7 @@ class Spec(object):
         # deref_flattened_spec depends on flattened_spec which assumes that model
         # discovery is performed
         run_post_processing(self)
+        self.resources = build_resources(self)
 
         if self.config['internally_dereference_refs']:
             deref_flattened_spec = self.deref_flattened_spec
@@ -253,6 +254,7 @@ class Spec(object):
             # Rebuild definitions using dereferences specs as base
             # this ensures that the generated models have no references
             run_post_processing(tmp_spec)
+            self.resources = build_resources(tmp_spec)
             self.definitions = tmp_spec.definitions
 
             # Avoid to evaluate is_ref every time, no references are possible at this time
@@ -262,7 +264,6 @@ class Spec(object):
             self.register_format(format)
 
         self.api_url = build_api_serving_url(self.spec_dict, self.origin_url)
-        self.resources = build_resources(self)
 
     @cached_property
     def _internal_spec_dict(self):
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -367,10 +367,7 @@ class Spec(object):
             self.build()
 
         return strip_xscope(
-            spec_dict=flattened_spec(
-                swagger_spec=self,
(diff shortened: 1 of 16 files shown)


bravado_core/spec_flattening.py
```diff
@@ -254,10 +254,11 @@ def descend(value):
             # Minimalistic swagger spec like object
             # it's not a valid spec due to lack of info and paths, but it's good enough to trigger model discovery
             spec_dict={
-                'definitions': {
+                known_mappings_key: {
```
sjaensch: How does this flattened spec actually look? Could you add a test that compares the flattened spec to what we expect it to look like, or at least post a gist with an example? It's not clear to me why we don't use `definitions` anymore...?

bravado_core/spec.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, spec_dict, origin_url=None, http_client=None,
             handlers=build_http_handlers(http_client),
         )
 
+        self._internal_spec_dict = spec_dict
```
sjaensch: Could you add a comment here that indicates why you're creating this attribute, and that it might be changed by run_post_processing? That information was really clear in the old implementation, but now is quite hidden.

bravado_core/spec.py
```diff
@@ -245,6 +246,7 @@ def run_post_processing(swagger_spec):
         # deref_flattened_spec depends on flattened_spec which assumes that model
         # discovery is performed
         run_post_processing(self)
+        self.resources = build_resources(self)
```
sjaensch: Could you add the info you provided in the description as a comment here, i.e. that flattening the spec requires resources to be present?

tests/spec/from_dict_test.py
```diff
@@ -125,3 +127,33 @@ def test_spec_with_dereffed_and_tagged_models_works(minimal_swagger_dict):
...
+        origin_url=get_url(multi_file_multi_directory_abspath),
+        config={'internally_dereference_refs': request.param},
+    )
+
+
+def test_flattened_multi_file_multi_directory_specs(multi_file_multi_directory_spec):
+    try:
+        multi_file_multi_directory_spec.flattened_spec
+    except BaseException as e:
+        pytest.fail('Unexpected exception: {e}'.format(e=e), e.__traceback__)
```
sjaensch: I'm still not happy with this test - it proves that there is no exception when flattening the spec, but tells us nothing about if the spec is correct. Could we not have a separate file (e.g. test-data/2.0/multi-file-multi-directory-spec/flattened_spec.yaml) and compare it to the result here?

I think it's fine to save whatever `flattened_spec` returns to that file; and as part of the review here, we can make sure the output is correct and equivalent to the multi-file spec.

test-data/2.0/multi-file-multi-directory-spec/flattened-multi-file-multi-directory-spec.json
```diff
@@ -0,0 +1,213 @@
+{
+  "consumes": [
+    "application/json"
+  ],
+  "definitions": {
+    "lfile:definitions.yaml": {
```
sjaensch: This is very helpful!

Note that the flattened spec you're creating is not a valid Swagger spec. Note that [according to the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#definitionsObject), the definitions object is a key/value mapping; the mapping being Schema objects. In your case, it's a mapping of a mapping of schema objects. Consequently, adding this spec to the [Swagger editor](https://editor.swagger.io/) produces a validation error. Similarly, swagger_spec_validator complains about this fact as well.

test-data/2.0/multi-file-multi-directory-spec/flattened-multi-file-multi-directory-spec.json
```diff
@@ -0,0 +1,213 @@
...
+      "get": {
+        "description": "This is the description of a random get endpoint",
+        "operationId": "endpoint_v1",
+        "parameters": [
+          {
+            "$ref": "#/parameters/lfile:endpoint..parameters.yaml|..parameters..number"
+          }
+        ],
+        "responses": {
+          "$ref": "#/definitions/lfile:endpoint..v1..responses.yaml"
```
sjaensch: You're not allowed to use references here. You'll need to specify all of the response HTTP codes inline (or at least the "default" key). The value can then be a reference object. Check the [spec definition of the `Responses` object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responsesObject). The Swagger editor catches this, swagger_spec_validator apparently does not.

macisamuele: It's really strange ... but you're definitely right ... I would have expected complains from swagger_spec_validator about it.
I'll open an issue on the library ... in our case in order to provide valid specs we need to remove the reference from [here](https://github.com/macisamuele/bravado-core/blob/maci-fix-flattening-failure-due-to-resolution-scope/test-data/2.0/multi-file-multi-directory-spec/endpoint/v1/paths.yaml#L7) as it is not correct. Starting from correct specs we'll obtain valid response

macisamuele:
* Improved heuristic to take into account the presence of `type: object` in order to return `ObjectType.SCHEMA`.
In order to achieve this, without changing the public interface I've added an optional parameter `default_type_to_object` that correspond to the equivalent value in the configs
* as we need the config value into `flattened_spec` I've added an extra parameter to it that allows to pass the whole `Spec` object rather than all it's components; the parameters list was growing too much
By doing this the calls that the library does on the function could be simplified without providing a backward incompatible change to the public interface. (we might think about adding a deprecation warning in the future to remove the long list of parameters)

bravado_core/spec_flattening.py
```diff
@@ -160,12 +160,20 @@ def flattened_spec(
     :param marshal_uri_function: function used to marshal uris in string suitable to be keys in Swagger Specs.
     :type marshal_uri_function: Callable with the same signature of ``_marshal_uri``
     :param spec_definitions: known swagger definitions (hint: definitions attribute of bravado_core.spec.Spec instance)
-    :type dict: bravado_core.spec.Spec
+    :type spec_definitions: dict
+    :param swagger_spec: bravado-core Spec object
+    :type swagger_spec: bravado_core.spec.Spec
```
sjaensch: I think we should document somewhere that you only really need to pass in swagger_spec for this function to work. Also, you do need either spec_dict or swagger_spec, right? You should add a check for that at the beginning of the function.

bravado_core/spec_flattening.py
```diff
@@ -161,18 +166,37 @@ def flattened_spec(
...
-        spec_definitions = swagger_spec.definitions if spec_definitions is None else spec_url
+        spec_dict = swagger_spec.spec_dict
+        spec_resolver = swagger_spec.resolver
+        spec_url = swagger_spec.origin_url
+        http_handlers = build_http_handlers(swagger_spec.http_client)
+        spec_definitions = swagger_spec.definitions
+
+    if spec_dict is None:
+        raise ValueError(
+            'spec_dict is None, the method assumes to receive a valid swagger spec dict'
```
sjaensch: either a valid swagger spec dict or a swagger spec object

PR #270, opened 2018-05-28
Refactory model discovery and model methods deprecation

macisamuele:
This PR is in preparation of the next bravado-core major release

Model discovery was defined half in `bravado_core/spec.py` and in `bravado_core/model.py` additionally was exposing _internal_ functions on the public interface.

This PR moves the methods related to model discovery into `bravado_core/model.py` and removes the public signatures of model collections models and `bravado_core.model.Model` methods that were traced as deprecated.

Changelog has been updated to make explicit the function signature update
Internal ref: COREBACK-5319

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -2,7 +2,6 @@
 import abc
 import functools
 import logging
-from warnings import warn
 
 import six
 from six import iteritems
@@ -478,13 +477,6 @@ class Model(object):
         )
         return model
 
-    def marshal(self):
-        warn(
-            "Model object methods are now prefixed with single underscore - use _marshal() instead.",
-            DeprecationWarning,
-        )
-        return self._marshal()
-
     def _marshal(self):
         """Marshal into a json-like dict.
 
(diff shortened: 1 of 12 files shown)


bravado_core/model.py
```diff
@@ -490,14 +485,6 @@ def _marshal(self):
         from bravado_core.marshal import marshal_model
         return marshal_model(self._swagger_spec, self._model_spec, self)
 
-    @classmethod
-    def unmarshal(cls, val):
```
sjaensch: The removal of `marshal` and `unmarshal` will break quite a bit of internal code - unlike the other methods, these ones are used in testing quite a bit. Additionally, the implementation is easy to maintain and there haven't been any reported issues with name clashes. Do you think it's worth removing them anyway?

macisamuele: I thought that could make sense to do it now as marked as deprecated.
But you're correct ... we heavily rely on it internally ... I'm ok taking a step back here

PR #282, opened 2018-06-29
Additional sanitization rules to cover more cases

sjaensch:
A few bugfixes around name sanitization; I ran into these cases when dealing with some of our internal specs.

1. We weren't stripping more than one leading digit (or underscore). We need to strip all digits.
1. In some cases the result can be an empty string; I've come up with a way to deal with it that works best when the name is mostly digits. There are still edge cases where this produces only an underscore, but I really hope those aren't real use cases. :)

### DIFF
--- a/bravado_core/util.py
+++ b/bravado_core/util.py
@@ -17,7 +17,17 @@ SANITIZE_RULES = [
     for regex, replacement in [
         ('[^A-Za-z0-9_]', '_'),  # valid chars for method names
         ('__+', '_'),  # collapse consecutive _
-        ('^[0-9_]|_$', ''),  # trim leading/trailing _ and leading digits
+        ('^[0-9_]+|_+$', ''),  # trim leading/trailing _ and leading digits
+    ]
+]
+
+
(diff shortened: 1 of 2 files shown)


bravado_core/operation.py
```diff
@@ -18,13 +18,13 @@ def _sanitize_operation_id(operation_id, http_method, path_name):
 
     # Handle crazy corner cases where someone explicitly sets operation
     # id a value that gets sanitized down to an empty string
-    if len(sanitized_operation_id) == 0:
+    if sanitized_operation_id in {'', '_'}:
```
macisamuele: Is `_` a possible case with the new changes in `sanitize_name`?

bravado_core/operation.py
```diff
@@ -18,13 +18,13 @@ def _sanitize_operation_id(operation_id, http_method, path_name):
 
     # Handle crazy corner cases where someone explicitly sets operation
     # id a value that gets sanitized down to an empty string
```
macisamuele: nit. please update the comment :)

PR #283, opened 2018-06-29
Refactor spec flattening

macisamuele:
The objective of this PR is to refactor the big `bravado_core.spec_flattening.flattened_spec` function into a more more testable/inspectable approach.

The strategy used consist of:
 * define a `_SpecFlattener` object that exposes the required functionalities
 * `flattened_spec` has been split into multiple methods that are kept of a reasonable length and separated by _competence_ area
 * all the small optimisations done to reduce attribute access are kept thanks to `@cached_property`


tests/model/model_discovery_test.py
```diff
@@ -28,6 +29,38 @@ def test_model_discovery_flow_with_ref_dereference(mock__run_post_processing, mi
...
+            },
+            json_reference='#/definitions/mock_model/x-model',
+            models=spec.definitions,
+            swagger_spec=swagger_spec,
+        )
+
+    # Side effect is needed to ensure that at least a model has been discovered before
+    # flattening process
+    mock__run_post_processing.side_effect = _run_post_processing_side_effect
+    model_discovery(swagger_spec=spec)
```
sjaensch: This indicates a flaw in design or in test setup. You're mocking out a function that's being called by the function you're testing, and then you're making sure that a variant of what that mocked out function does is actually executed. You should either:

- Not mock out _run_post_processing, or only mock out some of the methods it calls (and that can't run as part of your test)
- mock out _run_post_processing, but set the state accordingly in your test setup code; ideally you do this declaratively. If you're having trouble doing that cleanly it means that you're having issues with state handling; i.e. global or otherwise shared state that is difficult to manage in your test environment. This is one advantage of passing around state as arguments and return values.

As it stands right now it is impossible for me to understand what this test does exactly, and what the state before, during and after the execution of `model_discovery` is. In other words, whether this test passes or not doesn't tell me anything.

macisamuele: Fair point.
I changes the test to wrap `_run_post_processing` instead of mocking it.

Doing so allows `_run_post_processing` logic to be called and tests could still inspect how many times it was called

PR #286, opened 2018-07-03
Generalize parameter encoding (as done for parameter string decoding)

### LINKED ISSUE #281: Incorrect marshalling of boolean query params
Take for example this param:
```json
"parameters": [
          {
            "name": "force",
            "type": "boolean",
            "in": "query"
          }
]
```

Currently when bravado-core tries to marshal this param, it just casts it to `str()` which leads to `True/False` instead of `true/false`

Temporary ugly monkeypatch snippet for those, who use bravado+pytest:
```python
from bravado_core.param import marshal_param

def _marshal_param(param, value, request):
    marshal_param(param, value, request)
    if param.name in request['params'] and type(value) is bool:
        value = str(value).lower()
        request['params'][param.name] = value

monkeypatch.setattr('bravado.client.marshal_param', _marshal_param)
```

macisamuele:
The goal of this PR is to fix #281 

The approach used is really similar to the one already used for `unmarshal_param`.
I've defined a function that stringifies the paramater if its location is `header` or `query`

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -277,10 +277,10 @@ def encode_request_param(param_type, param_name, param_value):
         return None
 
     try:
+        param_value = str(param_value)
         if param_type == 'boolean':
-            return 'true' if param_value is True else 'false'
-        else:
-            return str(param_value)
+            param_value = param_value.lower()
+        return param_value
     except (ValueError, TypeError):  # pragma: no cover
         # Conversion to string should never fail, but as unicode issues are always
         # a thing in python I would rather prefer logging the issue instead of
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import logging
-from collections import defaultdict
 from functools import partial
 
 import simplejson as json
@@ -262,15 +261,6 @@ def cast_request_param(param_type, param_name, param_value):
         return param_value
 
 
-# Method to encode a parameter type to string
-_ENCODE_TYPE_TO_FUNC = defaultdict(
-    lambda: lambda x: str(x),
-    {
-        'boolean': lambda item: str(item).lower()
-    },
-)
-
-
 def encode_request_param(param_type, param_name, param_value):
     """
     Tries to cast a request param from its specified type in scheme to a string.
@@ -281,14 +271,16 @@ def encode_request_param(param_type, param_name, param_value):
     :param param_name: param name
     :type  param_name: string
     :param param_value: param value
-    :type  param_value: string
     """
     if param_value is None:  # pragma: no cover
         # We should never get into this branch, but better be more defensive
         return None
 
     try:
-        return _ENCODE_TYPE_TO_FUNC[param_type](param_value)
+        if param_type == 'boolean':
+            return 'true' if param_value is True else 'false'
+        else:
+            return str(param_value)
     except (ValueError, TypeError):  # pragma: no cover
         # Conversion to string should never fail, but as unicode issues are always
         # a thing in python I would rather prefer logging the issue instead of
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+from collections import defaultdict
 from functools import partial
 
 import simplejson as json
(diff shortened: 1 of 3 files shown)


bravado_core/param.py
```diff
@@ -260,6 +262,45 @@ def cast_request_param(param_type, param_name, param_value):
         return param_value
 
 
+# Method to encode a parameter type to string
+_ENCODE_TYPE_TO_FUNC = defaultdict(
+    lambda: lambda x: str(x),
+    {
+        'boolean': lambda item: str(item).lower()
+    },
```
sjaensch: I'd prefer adding this special case for boolean in the function below instead of providing this quite confusing defaultdict. 
I'm having a hard time finding where it specifies in the swagger spec how booleans should be represented as query or path parameters. I guess it's ok to lowercase the values, but I'd like to point out that this might be implementation-dependant (and thus lowercasing might not be "more correct" than leaving it as-is).

bravado_core/param.py
```diff
@@ -260,6 +262,45 @@ def cast_request_param(param_type, param_name, param_value):
...
+    except (ValueError, TypeError):  # pragma: no cover
+        # Conversion to string should never fail, but as unicode issues are always
+        # a thing in python I would rather prefer logging the issue instead of
+        # raising the exception
+        log.warning(
+            "Failed to encode %s value of %s from type %s to string",
+            param_name, param_value, param_type,
+        )
+        # Ignore type error, let jsonschema validation handle incorrect types
+        return param_value
```
sjaensch: What if people do not have request validation turned on? The code that is calling this function would throw a SwaggerMappingError, why did you opt not to do it here?

macisamuele: I basically inverted the logic in `cast_request_param` to perform the inverse conversion.
That's the reason for not throwing the exception

bravado_core/param.py
```diff
@@ -260,6 +262,45 @@ def cast_request_param(param_type, param_name, param_value):
...
+    :param param_type: name of the type to be casted to
+    :type  param_type: string
+    :param param_name: param name
+    :type  param_name: string
+    :param param_value: param value
+    :type  param_value: string
+    """
+    if param_value is None:  # pragma: no cover
+        # We should never get into this branch, but better be more defensive
+        return None
```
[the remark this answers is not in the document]
macisamuele: Fair point , I have to do double check
As side note: docstring does not look correct as `param_value` could have any type (even `bool` :) )

bravado_core/param.py
macisamuele: Sure, I proposed that approach to make easily extensible if needed
As we don't necessarily need it know I will "manually" handle the case

About _swagger spec how booleans should be represented as query_ I don't think that there is a specific guideline, but I would assume that most of the users will expect JSON-like booleans.
`bravado-core` has already implemented the sugar code needed to handle `True`/`true`/...

bravado_core/param.py
sjaensch: Yeah I've looked at that code, but you're not doing that really, are you? You just have a pretty non-obvious way to a) cast everything to `str`, and b) lowercase it afterwards if the type is `boolean`. I'd prefer that to be expressed in a more readable way. The code in param.py does that quite nicely; I don't think that structure works well the other way around. I don't think we need to follow that exact _structure_, just the logic.

macisamuele: Just to be sure: are you suggesting to remove the try-except statement?

sjaensch: I'm saying we should remove `_ENCODE_TYPE_TO_FUNC` unless there's a compelling reason for having it; the fact that `cast_request_param` has it doesn't mean it works well in this case. Most notably, `cast_request_param` didn't use `defaultdict` and three `lambda`s in the definition. :)
I think there's a much clearer way to write this in a simple `if` statement.

macisamuele: Cool
I got the point of removing the default dict, I was not sure about the exception handling.

Going to publish a new iteration soon

bravado_core/param.py
```diff
@@ -281,14 +271,16 @@ def encode_request_param(param_type, param_name, param_value):
...
     if param_value is None:  # pragma: no cover
         # We should never get into this branch, but better be more defensive
         return None
 
     try:
-        return _ENCODE_TYPE_TO_FUNC[param_type](param_value)
+        if param_type == 'boolean':
+            return 'true' if param_value is True else 'false'
+        else:
+            return str(param_value)
```
sjaensch: lg2m! Another way to do it:

    param_value = str(param_value)
    if param_type == 'boolean':
        param_value = param_value.lower()

bravado_core/param.py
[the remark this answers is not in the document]
sjaensch: I'm just wondering what the point of this function is; the name suggests it should return something suitable for sending a request, so ideally it should only return strings, right?

PR #295, opened 2018-09-13
Don't use Swagger spec path for API URL

mulmschneider:
According to the Swagger spec we should not use the path part of the
spec URL, but instead just use the host part of the URL (see the
existing comment for build_api_serving_url, as well as
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields
).

Also update tests to match new behaviour.

I don't know if bravado users depend on the current behavior - feedback appreciated.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -435,6 +435,6 @@ def build_api_serving_url(spec_dict, origin_url=None, preferred_scheme=None):
         return schemes[0]
 
     netloc = spec_dict.get('host', origin.netloc)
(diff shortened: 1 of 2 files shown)


mulmschneider:
Not sure why the 'docs' part of the build failed - seems to be unrelated from this PR?

DStape:
Hello, I've opened #299 because this change has caused some test failures for us, there's more detail in that thread, any feedback is appreciated, thanks.

Issue #299, opened 2018-09-20
path portion of 'origin_url' being ignored in build_api_serving_url()


DStape:
When an origin_url of http://localhost/x is specified, ```/x``` is expected to be present in all URLs that the function returns.

This was the behaviour until https://github.com/Yelp/bravado-core/commit/108363636068fc6b1234bad2a9139550e256af3a was added.

The message in that commit indicates the behaviour change was made because, "According to the Swagger spec we should not use the path part of the spec URL".

However, I'm not sure I agree. Looking at https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields, the only restriction I see is, "The basePath does not support path templating.".

In our use case, the basePath is a variable and we've been using origin_url in our bravado-based tests which are now broken with the change. So for now, we've pinned bravado-core to a lower version.

I'd like to see the original behaviour (full origin_url being honoured) re-instated, especially since I don't see any reason why this would violate the swagger 2.0 spec (although I may be mis-reading it, so please correct me :)).

@mulmschneider , are you able to expand on why the behaviour change was made?

Thanks for your time,
David

mulmschneider:
Hi there,

I was afraid that my change would create fallout for somebody, sorry it hit you.

I was referring to the following sentence in the API spec:
"If it is not included, the API is served directly under the host". 
Which I interpreted to mean that if `basePath` is not specified in the swagger.json you should indeed just use the host part of URL.

After re-reading the spec I'm not sure though if this still applies if you also don't specify the `host` field in your API spec (it is not a required field).

My problem that caused me to propose the change was the following piece of code:
```
from bravado.client import SwaggerClient                                                                                                                                                                                                                                              
client = SwaggerClient.from_url('https://172.30.251.16/swagger.json')  
```
Which, since my swagger.json didn't include a basePath, caused bravado to build API urls like `https://172.30.251.16/swagger.json/getpets` instead of  `https://172.30.251.16/getpets`.

Note that you can use the following to overwrite the auto generated URL:
```
client.swagger_spec.api_url = 'http://172.30.251.16/'
```


(Meta: I used the term `API spec` interchangeably for either the Swagger specification or the specification of the Swagger specification :/ - are there any better terms that clearly differentiate between the two?)

DStape:
No probs.

[...]

PR #300, opened 2018-09-21
Add new 'use_spec_url_for_base_path' boolean config option

DStape:
This new option addresses issues arising from 'basePath' not being
present in a given spec.

According to the 2.0 spec, "If it is not included, the API is served directly
under the host. The value MUST start with a leading slash (/).".

10836363 made the SwaggerClient compliant with the spec (basePath is now
implicitly set to '/' if field is missing from spec). However, this broke
existing code that utilizes the SwaggerClient (see Issue #299).

The current (spec-compliant) behaviour remains with this new option as
it defaults to False. Overriding this default with 'True' means the path
element of the URL used to retrieve the spec will be used instead.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -74,6 +74,12 @@ CONFIG_DEFAULTS = {
     # Completely dereference $refs to maximize marshaling and unmarshalling performances.
     # NOTE: this depends on validate_swagger_spec
     'internally_dereference_refs': False,
+
+    # What value to assume for basePath if it is missing from the spec (this
+    # config option is ignored if basePath is present in the spec)
+    # If True, use the 'path' element of the URL the spec was retrieved from
+    # If False, set basePath to '/' (conforms to Swagger 2.0 specification)
+    'use_spec_url_for_base_path': False
 }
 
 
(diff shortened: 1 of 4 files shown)


DStape:
UTs still to be updated, but I'd like some feedback first before spending time on those, just to make sure I'm not doing anything that's obviously silly, thanks.

DStape:
Gentle bump - @macisamuele , @sjaensch

bravado_core/spec.py
```diff
@@ -380,7 +388,9 @@ def read_file(uri):
     }
 
 
-def build_api_serving_url(spec_dict, origin_url=None, preferred_scheme=None):
+def build_api_serving_url(spec_dict, origin_url=None,
+                          use_spec_url_for_base_path=False,
```
macisamuele: I would recommend to add the new parameter at the end of the already existing parameters.
This method is _public_ and users could be using it passing `args` instead of `kwargs`, by adding `use_spec_url_for_base_path` before `preferred_scheme` we are providing a backward incompatible signature update

DStape: Fixed.

DStape:
Updated the doc string and added new tests.

DStape:
Fixed failing python3 test (due to urllib differences between python 2 and 3).

DStape:
Thanks @macisamuele.

DStape:
One last thing @macisamuele - is there an ETA on when this will be released? (Presumably v5.0.8?) Thanks again.

Issue #303, opened 2018-10-30
unmarshal_request breaks with jsonchema 3.x.x


gabisurita:
Apparently, bravado_core is trying to access a validator method that doesn't exists on jsonschema>=3.0.0a1. This version is within the range of valid dependencies: https://github.com/Yelp/bravado-core/blob/master/setup.py#L14

Here's my traceback:

```
―――――――――――――――――――― OpenAPIPluginsTest.test_delete_history ――――――――――――――――――――
self = <tests.openapi.test_plugins.OpenAPIPluginsTest testMethod=test_delete_history>
    def test_delete_history(self):
        op = self.resources["History"].delete_history
        self.request.path = {"bucket_id": "b1"}
>       self.validate_request_call(op)
/home/travis/build/Kinto/kinto/tests/openapi/test_plugins.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/build/Kinto/kinto/tests/openapi/support.py:94: in validate_request_call
    params = unmarshal_request(self.request, op)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/bravado_core/request.py:66: in unmarshal_request
    param_value = unmarshal_param(param, request)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/bravado_core/param.py:200: in unmarshal_param
    validate_schema_object(swagger_spec, param_spec, raw_value)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/bravado_core/validate.py:53: in validate_schema_object
    validate_primitive(swagger_spec, schema_object_spec, value)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/bravado_core/validate.py:25: in scrubbed
    return func(*args, **kwargs)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/bravado_core/validate.py:79: in validate_primitive
    resolver=swagger_spec.resolver).validate(value)
/home/travis/build/Kinto/kinto/.tox/py35/lib/python3.5/site-packages/jsonschema/validators.py:304: in validate
    for error in self.iter_errors(*args, **kwargs):
[...]

PR #304, opened 2018-10-30
Remove direct usage of jsonschema._validators

macisamuele:
Fixes: #303 

The goal of this PR is to guarantee bravado-core interoperability with jsonschema version 3 (released only in alpha for now).
The new release is changing how the validator methods are named within jsonschema._validators package.
In general we should not _import_ private definitions, but this was done.

In order to make bravado-core able to work with the current jsonschema releases and the new release (alpha released) we need to call the _original_ validators using a different approach.
Instead of calling validator methods from the private module we could extract them from `Draft4Validator.VALIDATORS` class attribute (by using only public informations).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,7 @@
--   repo: git://github.com/pre-commit/pre-commit-hooks
-    sha: v0.7.1
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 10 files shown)


macisamuele:
Updated pre-commit hooks and fixed build (that was failing also on master).

gabisurita:
@macisamuele Thank you! Do you know when you expect to release this fix?

PR #204, opened 2017-10-16
Fully dereference specs for performance improvements

macisamuele:
The goal of this PR is to enhance bravado-core performances by using fully dereferenced specs for marshaling and unmarshaling operations.

According to how bravado-core is structured all the decisions made during marsharling and unmarshaling are made on top of ``spec.client_dict`` object which could contain references (``{"$ref": ...}`` objects). In order to make the spec traversing as transparent as possible we take advantage of ``spec.deref`` method which:
 - checks if the object is a potential reference object
 - eventually dereference the reference

Dereferencing the reference is _expensive_ especially considering that this will be done for all the unmarshaling (no specific caches are used).
A possible solution is to completely dereference the spec at build time, doing so ``deref`` method will always return the input object since no reference objects are possible at that point.

ℹ️  ``internally_dereference_refs`` is defaulted to ``False`` and a warning is rendered in case this config is used. I don't think that this will be needed so I'm fine to remove it 😄 

⚠️  currently are not present tests for checking ``deref_flattened_spec`` -> I'll write the test soon ... in the mean time any type of feedback is welcome

The performance improvements (based on ``tests/profiling/unmarshal_response_profiler_test.py``) are estimated around 35-40%.

```
------------------------------------------------------------------------------------- benchmark 'number_of_objects=100': 8 tests ------------------------------------------------------------------------------------
Name (time in ms)                                       Min                Max               Mean            StdDev             Median                IQR            Outliers       OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_small_objects[not_validate-full-deref-100]      3.4703 (1.0)       7.4356 (1.0)       6.6310 (1.0)      0.4756 (1.63)      6.5699 (1.0)       0.2250 (1.0)         47;49  150.8076 (1.0)         209           1
test_small_objects[validate-full-deref-100]          7.4766 (2.15)     14.2583 (1.92)     12.5795 (1.90)     1.6259 (5.58)     12.8922 (1.96)      2.3475 (10.44)        15;2   79.4944 (0.53)         72           1
test_small_objects[not_validate-with-refs-100]       9.5869 (2.76)     16.3894 (2.20)     14.8315 (2.24)     1.0073 (3.46)     15.0546 (2.29)      0.2467 (1.10)        15;27   67.4241 (0.45)        105           1
[...]

PR #305, opened 2018-11-19
Unmarshal array inline

Timothyyung:
Swapped the call to unmarshal-schema-object in unmarshal array to a direct call. 
Also Reduced the calls to deref only dereffing one time for n object ( Removing the redundant function calls )

### DIFF
--- a/bravado_core/unmarshal.py
+++ b/bravado_core/unmarshal.py
@@ -64,7 +64,7 @@ def unmarshal_schema_object(swagger_spec, schema_object_spec, value):
         return value
 
     raise SwaggerMappingError(
-        "Don't know how to unmarshal value {0} with a type of {1}"
+        "Don't know how somthing to unmarshal value {0} with a type of {1}"
         .format(value, obj_type))
 
 
@@ -94,7 +94,7 @@ def unmarshal_array(swagger_spec, array_spec, array_value):
     :type array_value: list
     :rtype: list
     :raises: SwaggerMappingError
-    """
+    )"""
     if array_value is None:
         return handle_null_value(swagger_spec, array_spec)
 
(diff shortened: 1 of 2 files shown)


bravado_core/unmarshal.py
```diff
@@ -64,7 +64,7 @@ def unmarshal_schema_object(swagger_spec, schema_object_spec, value):
         return value
 
     raise SwaggerMappingError(
-        "Don't know how to unmarshal value {0} with a type of {1}"
+        "Don't know how somthing to unmarshal value {0} with a type of {1}"
```
macisamuele: with this addition the exception message seems not correct (at least from an english prospective)
could you please revert this change?

bravado_core/unmarshal.py
```diff
@@ -94,7 +94,7 @@ def unmarshal_array(swagger_spec, array_spec, array_value):
     :type array_value: list
     :rtype: list
     :raises: SwaggerMappingError
-    """
+    )"""
```
macisamuele: please remove it

bravado_core/unmarshal.py
```diff
@@ -103,10 +103,65 @@ def unmarshal_array(swagger_spec, array_spec, array_value):
             type(array_value), array_value))
 
     item_spec = swagger_spec.deref(array_spec).get('items')
-    return [
-        unmarshal_schema_object(swagger_spec, item_spec, item)
-        for item in array_value
-    ]
+
+    # deref objects outside of the loop and pass unmarshal without going through 'unmarshal_schema_object'
```
macisamuele: I'm not a big fan of this approach as this is basically inlining the logic of `unmarshal_schema_object` in `unmarshal_array`.
Doing so will increase the likelihood of providing inconsistent changes if we're going to modify/update `unmarshal_schema_object` implementation.

I see that the main objective is to reduce the number of calls to `swagger_spec.deref`.
I would suggest two alternative approaches to this:
1) have you tried to use `internally_dereference_refs` configuration? Thanks to the flag all the specs are dereferenced once and `swagger_spec.deref` will be a no-op (Check #204 for more details)
2) (if point 1 does not help) you could wrap `Spec._force_deref` with `bravado_core.util.memoize_by_id` to make `swagger_spec.deref` almost a no-op after the first call

macisamuele:

PR #306, opened 2018-11-20
Memoization

albert831229:
Memoize the result of deref(), collapse_properties(), get_spec_for_prop() function

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,10 @@
 Changelog
 =========
+5.0.8 (2018-11-19)
+------------------
+- Memoize the result of deref() function in Spec.py, collapse_properties(), get_spec_for_prop() in Schema.py
+- Add a dictionary in Spec.py to store the result
+- Change all deref fuction called in unmarshalling process to fast_deref
 
 5.0.7 (2018-09-18)
 ------------------
--- a/bravado_core/schema.py
+++ b/bravado_core/schema.py
@@ -180,34 +180,3 @@ def collapsed_properties(model_spec, swagger_spec):
             properties.update(more_properties)
 
     return properties
-
-
-@memoize_by_id
-def fast_collapsed_properties(model_spec, swagger_spec):
-    """Processes model spec and outputs dictionary with attributes
-    as the keys and attribute spec as the value for the model.
-
-    This handles traversing any polymorphic models and the hierarchy
-    of properties properly.
-
-    :param model_spec: model specification (must be dereferenced already)
-    :type model_spec: dict
-    :param swagger_spec: :class:`bravado_core.spec.Spec`
-    :returns: dict
-    """
-    properties = {}
-
-    # properties may or may not be present
-    if 'properties' in model_spec:
-        for attr, attr_spec in iteritems(model_spec['properties']):
-            properties[attr] = attr_spec
-
-    # allOf may or may not be present
-    if 'allOf' in model_spec:
-        deref = swagger_spec.deref
-        for item_spec in model_spec['allOf']:
-            item_spec = deref(item_spec)
-            more_properties = collapsed_properties(item_spec, swagger_spec)
-            properties.update(more_properties)
-
-    return properties
--- a/bravado_core/schema.py
+++ b/bravado_core/schema.py
@@ -97,39 +97,47 @@ def get_spec_for_prop(swagger_spec, object_spec, object_value, prop_name, proper
     :return: spec for the given property or None if no spec found
     :rtype: dict or None
     """
-    deref = swagger_spec.deref
+    i = id(prop_name)
+    try:
+        return swagger_spec.cache_schema[i]
+    except KeyError:
+        deref = swagger_spec.fast_deref
 
-    if properties is None:
-        properties = collapsed_properties(deref(object_spec), swagger_spec)
-    prop_spec = properties.get(prop_name)
+        if properties is None:
+            properties = collapsed_properties(deref(object_spec), swagger_spec)
+        prop_spec = properties.get(prop_name)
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -1,5 +1,10 @@
 Changelog
 =========
+5.0.8 (2018-11-19)
```
macisamuele: Thanks a lot for updating the changelog.
A couple of notes though:
 * please don't set the version and release date as is not said that a new version will be immediately released if the PR is merged. Having a version and release date set could make [documentation](https://bravado-core.readthedocs.io/en/stable/changelog.html) misleading
 * try to use the _style_ used by the other changelog items (one line per change and PR reference at the end of it)

bravado_core/schema.py
```diff
@@ -156,10 +154,8 @@ def handle_null_value(swagger_spec, schema_object_spec):
 def collapsed_properties(model_spec, swagger_spec):
     """Processes model spec and outputs dictionary with attributes
     as the keys and attribute spec as the value for the model.
-
```
macisamuele: nit. keep this line

bravado_core/schema.py
```diff
@@ -156,10 +154,8 @@ def handle_null_value(swagger_spec, schema_object_spec):
...
     This handles traversing any polymorphic models and the hierarchy
     of properties properly.
-
```
macisamuele: nit. keep this line

bravado_core/schema.py
```diff
@@ -182,3 +178,39 @@ def collapsed_properties(model_spec, swagger_spec):
             properties.update(more_properties)
 
     return properties
+
+
+def fast_collapsed_properties(model_spec, swagger_spec):
```
macisamuele: I would suggest to avoid re-implementing `collapsed_properties` but rather wrap it with `bravado_core.util.memoize_by_id` which already offers you this functionality.
Doing so you'll have:
 * way smaller code change
 * reduced likelihood to have `collapsed_properties` and `fast_collapsed_properties` out of sync
 * achieve your goal

bravado_core/schema.py
```diff
@@ -182,3 +178,39 @@ def collapsed_properties(model_spec, swagger_spec):
...
+    :returns: dict
+    """
+
+    i = id(model_spec)
+    try:
+        return swagger_spec.cache_schema[i]
+    except KeyError:
+        properties = {}
+
+    # properties may or may not be present
```
macisamuele: nit. comment is not aligned with the code

bravado_core/schema.py
```diff
@@ -182,3 +178,39 @@ def collapsed_properties(model_spec, swagger_spec):
...
+        if 'properties' in model_spec:
+            for attr, attr_spec in iteritems(model_spec['properties']):
+                properties[attr] = attr_spec
+
+    # allOf may or may not be present
```
macisamuele: nit. comment is not aligned with the code

bravado_core/spec.py
```diff
@@ -129,6 +129,8 @@ def __init__(self, spec_dict, origin_url=None, http_client=None,
         # spec dict used to build resources, in case internally_dereference_refs config is enabled
         # it will be overridden by the dereferenced specs (by build method). More context in PR#263
         self._internal_spec_dict = spec_dict
+        self.cache_spec = {}
+        self.cache_schema = {}
```
macisamuele: I personally don't like much adding instance properties for caching, I would rather prefer wrapping with decorators the function that I would like have cached.

By using variables like this is more complicated understanding what content to expect in the _cache_, who is writing in the cache, ensuring that all the writers write with the same semantic, etc. .
In this specific case we are saving into `bravado_core.spec.Spec.cache_schema` data  evaluated by `bravado_core.schema.fast_collapsed_properties` and the mapping is not really easy to debug (as you're saving `id`s instead of the model spec that you have available).

Additionally not prefixing the attribute name with `_` makes this "public", which means that we cannot remove it if we think that the benefits of this caching are not balanced by the costs.

bravado_core/spec.py
```diff
@@ -228,8 +230,25 @@ def _force_deref(self, ref_dict):
             _, target = self.resolver.resolve(ref_dict['$ref'])
             return target
 
+    def fast_deref(self, ref_dict):
```
macisamuele: This change could be replaced by wrapping with `bravado_core.util.memoize_by_id` the `bravado_core.spec.Spec._force_deref` function.

By applying the related changes to use `swagger_spec.fast_deref` instead of `swagger_spec.deref` you will reduce the dereferencing speed in case `internally_dereference_refs` config is in use. This happens because if `internally_dereference_refs` is `True` dereferencing is a no-op while with this implementation it will become an item access in a dictionary (that is slower than returning the input value).

CHANGELOG.rst
albert831229: Sorry for the late reply. Thank you so much for the advice! I removed the version and date, also modified the changes.

bravado_core/schema.py
albert831229: Add the line

bravado_core/schema.py
albert831229: Deleted all fast_* function and put memoize_by_id decorator before collapsed_properties and deref

bravado_core/spec.py
albert831229: Does this mean that memoizing the result of deref is not a good idea even if I wrap it with  "memoize_by_id" decorator?

albert831229:
I made a few Swagger specs and 1000kb JSON data  to test out unmarshal functionality. The performance increases about 20% when the JSON data is array of objects or nested array.

macisamuele:

PR #307, opened 2018-11-20
change unmarshal array

PeilinzhongZ:


bravado_core/unmarshal.py
```diff
@@ -32,7 +32,42 @@ def unmarshal_schema_object(swagger_spec, schema_object_spec, value):
...
+
+    if obj_type == 'file':
+        return value
+
+    raise SwaggerMappingError(
+        "Don't know how to unmarshal value {0} with a type of {1}"
+        .format(value, obj_type))
+
 
+def unmarshal_schema_object_sub(swagger_spec, schema_object_spec, value):
```
macisamuele: The implementation of this function looks like a copy-paste of `unmarshal_schema_object` without the initial dereferencing.
Why is this needed?

PeilinzhongZ: I am trying to minimize the number of call of deref(). Since unmarshal_array would using the same item_spec to call unmarshal_schema_object, copy-paste of it can reduce the redundant call of deref().

macisamuele: I see that you're trying to reduce the number of calls to `swagger_spec.deref`, but I don't see the reason for that other than performances.

If performances is your leading reason I would suggest to check https://github.com/Yelp/bravado-core/pull/305#discussion_r234901332 as it is really similar to what you are doing here.

The base point is that to improve performances you are duplicating a lot of code and this makes the code-base harder to maintain as changes in `unmarshal_schema_object` will not be reflected on the code that you copy-pasted and this could lead to inconsistent/surprising behaviour.

I would suggest to cache the `swagger_spec.deref` method (actually it's internal implementation provided in `bravado_core.spec.Spec._force_deref`).
If you do so you'll not reduce the number of calls to `deref` but you could make them faster.

If that will be your direction please make sure to post benchmarks that motivate how caching is helping you there. This is because caching has a cost, the memory usage will increase.

macisamuele:

PR #313, opened 2019-01-09
Normalise URIs

macisamuele:
While investigating internal issues I did noticed that the result of spec flattening is dependent on how the references looks like.

Specs like
```yaml
swagger: '2.0'
info:
  title: Test
  version: '1.0'
definitions:
  text:
    type: string
paths:
  /endpoint:
    get:
      responses:
        '200':
          description: ''
          schema:
            $ref: '#/definitions/text'
        '300':
          description: ''
          schema:
            $ref: '#definitions/text'
```
would be flattened in something like
```yaml
definitions:
  'lfile:.|..definitions..text':
    type: string
  'lfile:.|definitions..text':
    type: string
info:
  title: Test
  version: '1.0'
paths:
  /endpoint:
    get:
      responses:
        '200':
          description: ''
          schema:
            $ref: '#/definitions/lfile:.|..definitions..text'
        '300':
          description: ''
          schema:
            $ref: '#/definitions/lfile:.|definitions..text'
swagger: '2.0'
```

instead of
```yaml
definitions:
  'lfile:.|..definitions..text':
    type: string
info:
  title: Test
  version: '1.0'
paths:
  /endpoint:
    get:
      responses:
        '200':
          description: ''
          schema:
            $ref: '#/definitions/lfile:.|..definitions..text'
        '300':
          description: ''
          schema:
            $ref: '#/definitions/lfile:.|..definitions..text'
swagger: '2.0'
```

Note the duplicate definition of `text` (as `lfile:.|..definitions..text` and `lfile:.|definitions..text`)

This is caused by the fact that the original spec have two equivalent but not equal references to `text` (`#/definitions/text` and `#definitions/text`).

The goal of this PR is to ensure that once we're building a `bravado_core.spec.Spec` object we're performing some form of minimal normalisation of URIs such that duplication of definitions is reduced in case of flattening

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -15,6 +15,7 @@ from bravado_core.schema import is_list_like
 from bravado_core.schema import is_ref
 from bravado_core.schema import SWAGGER_PRIMITIVES
 from bravado_core.util import determine_object_type
+from bravado_core.util import normalize_uri
 from bravado_core.util import ObjectType
 from bravado_core.util import strip_xscope
 
@@ -707,9 +708,8 @@ def _post_process_spec(spec_dict, spec_resolver, on_container_callbacks):
             if json_reference is None:
                 json_reference = '{}#'.format(spec_resolver.resolution_scope)
 
-            is_reference = is_ref(fragment)
-            if is_reference:
-                ref = fragment['$ref']
+            if is_ref(fragment):
+                fragment['$ref'] = ref = normalize_uri(fragment['$ref'])
                 attach_scope(fragment, spec_resolver)
                 with spec_resolver.resolving(ref) as target:
                     if id(target) in cache:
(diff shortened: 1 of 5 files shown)


sjaensch:
I don't think what you're doing here is semantically correct. That first slash has a meaning: it identifies the value as a JSON pointer ([RFC 6901](https://tools.ietf.org/html/rfc6901). JSON schema does define the possibility to define relative references without using JSON pointers, by using `id` ([documentation](https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref) - note that Swagger 2 is using JSON Schema draft v4, where `$id` was still called `id`).

So what should happen in this case is that we should raise an error for the value `#definitions/text`, probably by saying that we can't find that reference.

macisamuele:
Going to drop this for now.
I'm not 100% sure about the change after more carefully reading the JSON pointer RFC

PR #314, opened 2019-01-10
Add typing support

macisamuele:
This is a WIP PR with branch based on #313 (so `normalise_uri` changes will be reviewed elsewhere)

I've added typing to few files where typing was really straight forward and on `Spec` mostly because it will be required as soon as `request`, `response`, `resource`, `operation` and basically everything will be typed.

Next steps will be to provide typing information to `bravado_core.response` but I would like to split it in multiple PRs as this looks already big 😃


bravado_core/security_requirement.py
```diff
@@ -2,9 +2,16 @@
...
 import six
+import typing
 
 from bravado_core.exception import SwaggerSchemaError
 
+if typing.TYPE_CHECKING:
+    from bravado_core.security_definition import SecurityDefinition  # noqa: F401
+    from bravado_core.spec import Spec  # noqa: F401
+
+    T = typing.TypeVar('T')
```
sjaensch: Where in the file are you using `T`?

bravado_core/spec.py
```diff
@@ -33,9 +39,16 @@
...
-
 log = logging.getLogger(__name__)
 
+if typing.TYPE_CHECKING:
+    T = typing.TypeVar('T', bound='Spec')
+
+    class HttpClientProtocol(typing.Protocol):
+        def request(self, request_params):
+            # type: (typing.MutableMapping[str, typing.Any]) -> typing.Any
+            pass
```
sjaensch: That seems kind of weak as a protocol for our HTTP clients? We might not want to introduce a circular dependency to bravado, but instead we could either move the HttpClient base class to bravado-core or change this protocol to more closely match reality. In particular:

- specify all parameters to `request`, also the optional ones
- correctly specify the return type (an instance of HttpFuture, which is in bravado-core)

Otherwise this protocol has the potential to break downstream type checking (i.e. for users of bravado/bravado-core).

bravado_core/model.py
```diff
@@ -707,9 +708,8 @@ def wrapper(fragment, json_reference=None):
             if json_reference is None:
                 json_reference = '{}#'.format(spec_resolver.resolution_scope)
 
-            is_reference = is_ref(fragment)
-            if is_reference:
-                ref = fragment['$ref']
+            if is_ref(fragment):
+                fragment['$ref'] = ref = normalize_uri(fragment['$ref'])
```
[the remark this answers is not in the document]
sjaensch: I now re-read the description and saw your mention of `normalize_uri`. Was there no way to separate the two? Either way I'm going to ignore those changes. :)

macisamuele:
This PR has to be rebased on master as #313 has been dropped.
Additionally we need to identify a valid approach to deal with the `HttpClientProtocol`.
Currently the best approach might be defining the protocol as an abstract class in `bravado_core` and update `bravado` to use it as base class for

https://github.com/Yelp/bravado/blob/384db8ae6e3950edc7cfeb2ba68c50273ceedcd7/bravado/http_client.py#L12-L15

macisamuele:
I'm closing this PR as a lot of progress has been done already on the codebase and as it is in a very un-mergeable state.

I'll be providing a new PR with typing for almost all the modules where the most complex ones (ie. where in this PR was defined the typing.Protocol) are not included such that we can provide some typing information earlier.

PR #317, opened 2019-02-11
Feature/slow binary handling

greenphire-greg:
This PR adds an optimization for binary responses based on content type. [This was originally documented as Issue 394 in the bravado repository](https://github.com/Yelp/bravado/issues/394).

### DIFF
--- a/bravado_core/content_type.py
+++ b/bravado_core/content_type.py
@@ -1,3 +1,8 @@
 # -*- coding: utf-8 -*-
+APP_EXCEL = 'application/vnd.ms-excel'
+APP_EXCEL_OPENXML = (
+    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+)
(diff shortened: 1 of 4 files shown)


macisamuele:
@greenphire-greg thanks a lot for contributing.
NOTE: this is a quick first pass (did not check all the details)

Checking at the issue and at your PR I have a couple of suggestions
 * bravado unmarshaling is performed [here](https://github.com/Yelp/bravado/blob/master/bravado/http_future.py#L342) and not in bravado-core (probably the bravado-core is some _dead_ code that we should eventually remove)
 * I would suggest to avoid to mark all the possible mime types that reflect binary formats as this is not scalable.

greenphire-greg:
Alright, I'll close this and resubmit to the bravado repo.

macisamuele:
@greenphire-greg thanks for the understanding and for helping us in making the tool better

PR #320, opened 2019-03-05
Fix Python 3.8 issue, use separate flake8 pre-commit hook, fix tests

sjaensch:
Starting with Python 3.8 it's not possible to import `Mapping` from `collections`, we'll need to use `collections.abc` as package.

Additionally, I'm introducing the separate flake8 pre-commit hook as well as changing things around pre-commit hook installation slightly to make sure running `tox` directly also installs them. I've also incorporated the fix from #319 with one slight modification so the tests pass on Python 3.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -3,12 +3,10 @@ import json
 import logging
 import os.path
 import warnings
-from contextlib import closing
 
 import yaml
 from jsonref import JsonRef
 from jsonschema import FormatChecker
-from jsonschema.compat import urlopen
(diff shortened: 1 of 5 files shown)


bravado_core/spec.py
```diff
@@ -375,7 +373,7 @@ def download(uri):
             return response.json()
 
     def read_file(uri):
-        with closing(urlopen(uri)) as fp:
+        with open(urlparse(uri).path, mode='rb') as fp:
```
macisamuele: Lovely change :)
I can drop my PR if this will be merged

PR #325, opened 2019-04-09
Implement deepcopy support for Model instances

sjaensch:
It was reported internally that deepcopy doesn't work on model instances and fails with an infinite recursion exception. The test I'm adding failed without my change.

In doing a deep copy we should only copy the model properties, not the metadata set on the class. All of them are stored in the `__dict` attribute.

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -3,6 +3,7 @@ import abc
 import functools
 import logging
 import re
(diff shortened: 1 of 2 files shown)


shrayus:
Looks great! Thanks for fixing this so quickly :) Shipit!

PR #330, opened 2019-05-20
Use lazy attributes on model

macisamuele:
The goal of this PR is to use lazy evaluated and cached attributes on the model class.
By doing so we can avoid to pre-evaluate, for all the models, certain features like `_properties` or `_inherits_from` which could be time consuming.

Thanks to the usage of `@lazy_class_attribute` we do evaluate the property only once per model type and reuse the evaluation result for the following calls.

An other change done by this PR is to use `slots` in the `Model`, thanks to this we can reduce the memory footprint of the model instances without negatively impacting the performances.

NOTE: PR #329 objective was only to extend `@cached_property` to be used in meta classes, the extension has not be used here because we would have had to override `__getattribute__` as well causing a performance hit of ~7-9%.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,40 @@
-.PHONY: all install test tests clean docs benchmark
 
+.PHONY: all
 all: test
 
+.PHONY: build
 build:
 	./setup.py bdist_egg
 
+.PHONY: dev
 dev: clean
 	./setup.py develop
 
+.PHONY: docs
 docs:
 	tox -e docs
 
+.PHONY: install
 install:
 	pip install .
 
+.PHONY: test
 test:
 	tox -- tests --ignore tests/profiling
 
+.PHONY: tests
 tests: test
+	@true
 
-benchmark: install-hooks
+.PHONY: benchmark
+benchmark:
 	tox -e benchmark
 
 .PHONY: install-hooks
 install-hooks:
 	tox -e pre-commit
 
+.PHONY: clean
 clean:
 	@rm -rf .benchmarks .tox build dist docs/build *.egg-info
 	find . -name '*.pyc' -delete
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -234,7 +234,7 @@ class ModelMeta(abc.ABCMeta):
         if not isinstance(type(instance), type(self)):
             return False
 
-        if not hasattr(self, '_model_spec'):
+        if type(self) == ModelMeta:
             # This is the base Model class
             return True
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1,31 +1,40 @@
...
 test:
 	tox -- tests --ignore tests/profiling
 
+.PHONY: tests
 tests: test
+	@true
 
-benchmark: install-hooks
+.PHONY: benchmark
+benchmark:
```
sjaensch: Any reason we're not installing hooks before running benchmarks anymore?

bravado_core/model.py
```diff
@@ -280,14 +281,36 @@ class Model(object):
...
-        property names to their specs. See
+        Class attribute that must be assigned on subclasses.
+        Dict mapping property names to their specs. See
         :func:`bravado_core.schema.collapsed_properties`.
+
+    .. attribute:: _inherits_from
+
+        Class attribute that must be assigned on subclasses.
+        List of the models from which the current model inherits from.
+        (The list will be not empty only for schemas with allOf)
```
sjaensch: The list will be non-empty only for schemas with allOf

bravado_core/model.py
```diff
@@ -280,14 +281,36 @@ class Model(object):
...
+
+
+    .. attribute:: _deny_additional_properties
+
+        Class attribute that must be assigned on subclasses.
+        Flag that specifies if the model can or cannot allow additional
```
sjaensch: does or does not allow

bravado_core/model.py
```diff
@@ -586,19 +642,10 @@ def create_model_type(swagger_spec, model_name, model_spec, bases=(Model,), json
...
-            if inherited_name:
-                inherits_from.append(inherited_name)
-
     return type(str(model_name), bases, dict(
         __doc__=ModelDocstring(),
         _swagger_spec=swagger_spec,
         _model_spec=model_spec,
-        _properties=collapsed_properties(model_spec, swagger_spec),
-        _inherits_from=inherits_from,
         _json_reference=json_reference,
```
sjaensch: Note that you'll need to declare `__slots__` on the actual model types as well if you want the benefit. I'm not sure if that would be a breaking change since we do override `__setattr__`.

Makefile
macisamuele: Running `test` and `benchmark` does not require having hooks installed.
The dependency was already removed from `test` but forgot to remove on `benchmark`

bravado_core/model.py
macisamuele: Not 100% sure about the possible breakage, tests are green and size wise we have a benefit, so I'm inclined to think that `slots` are working.
The fact that we override `__setattr__` removes the potential performance gain (removing a dictionary lookup) from the operation, but we should be covered in terms of interfaces

PR #331, opened 2019-05-21
Enhance unmarshal performances

macisamuele:
The goal of this PR is, as mentioned in the title, unmarshaling performance enhancement.
NOTE: A similar PR will be published for applying the same approach described here for the marshaling process. I would like to have this verified first in order to avoid having two massive branches open at the same time

In order to achieve unmarshaling performance boost I've _rewritten_ the whole `bravado_core.unmarshal` module.

The current, before this PR, logic of unmarshaling is:
 1. extract the type from the schema
 2. if the schema if of type x then call the appropriate unmarshaling function
     * if the schema is of type object: call the top unmarshaling function for the property (point 1)
     * if the schema is of type object: call  the top unmarshaling function for each item of the list (point 1)
     * if the schema is of primitive type: identify, each time, the right formatting function to use
 3. once the unmarshaling function is identified then perform the real unmarshaling

This PR aims to cache results of point 1 and 2 for later "cost-free" usage.

While making this changes I've noticed that there might be no real good reason to have all the unmarshaling functions public, users of the library commonly use `bravado_core.unmarshal.unmarshal_schema_object`.
In order to simplify future maintenance work I've marked all the other functions as deprecated in the next major release. Doing so will allows us to modify more, if needed, the internals of unmarshaling without worrying about breaking backward compatibility.

Be aware that:
 * `bravado_core.unmarshal.get_unmarshaling_method` is wrapped by `_decorators.wrap_recursive_call_exception` and `memoize_by_id` decorators.
    `_decorators.wrap_recursive_call_exception` decorator is needed as models could be recursive and so invoking `get_unmarshaling_method` might end up on an infinite recursion.
    In order to proactively deal with it I've modified `@memoize_by_id` decorator to raise a controlled exception if an unbounded recursion is identified (ℹ️ the interpreter would raise anyway with a `RuntimeError: maximum recursion depth exceeded`)
    `memoize_by_id` decorator is needed to speed up evaluation of the marshaling method for the same spec and schema. This effectively is one of the major performance boost (basically the expensive point 1 and 2 of the original flow become, more or less, a dictionary lookup)
 * ~unmarshaling of a polymorphic object does not raise the traditional `SwaggerMappingError`.
    ℹ️ : The change is intentional as we should not do validation if validation is disabled (and as it was it was not possible to enable/disable that validation)~ (addressed by #333 )
 * ~Add support for unmarshaling additionalProperties (it was not done previously) -> `test_unmarshal_object_with_additional_properties` test~  (addressed by #333 )
---------------
Let's get to the juicy part 😄 
Raw data on: https://gist.github.com/macisamuele/9ff23d3e5e81c40cd884425a7eae5751

This PR provides a massive performance improvement:
 * the worst performance improvement is ~46% (if in the branch takes 1 then on master takes 1.87)
 * the best performance improvement is ~70% (if in the branch takes 1 then on master takes 3.33)
 * the average performance improvement is ~60% (if the branch takes 1 then on master takes 2.57)

### DIFF
--- a/bravado_core/_compat.py
+++ b/bravado_core/_compat.py
@@ -5,3 +5,9 @@ if six.PY2:  # pragma: no cover  # py2
     from functools32 import wraps  # noqa: F401
 else:  # pragma: no cover  # py3+
     from functools import wraps  # noqa: F401
+
+
+try:
+    from collections.abc import Mapping  # noqa: F401  # pragma: no cover  # py3.3+
+except ImportError:
+    from collections import Mapping  # noqa: F401  # py3.2 or older
--- a/bravado_core/_decorators.py
+++ b/bravado_core/_decorators.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""
+The module defines useful decorators for the marshaling and unmarshaling process
+
+NOTE: The module has been marked as PRIVATE., so we are required to preserve
+      This is done to allow future modification to the defined decorators
+      without worrying about backward incompatible changes.
+"""
+from bravado_core import schema
+from bravado_core._compat import wraps
+from bravado_core.exception import SwaggerMappingError
+from bravado_core.util import memoize_by_id
+from bravado_core.util import RecursiveCallException
+
+
+@memoize_by_id
+def handle_null_value(swagger_spec, object_schema, is_nullable=False):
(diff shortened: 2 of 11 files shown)


macisamuele:
I'm closing this PR as the history is a bit confusing. A new PR, with the same changes, will be created and linked to this

PR #333, opened 2019-05-22
Enhance unmarshaling of additional properties and defaults

macisamuele:
The goal of the PR is to fix issues spotted while working on #331.

Namely I noticed that:
 * unmarshaling is not unmarshaling additional properties.
    Unmarshaling `{"key1": "2019-05-22"}` against the following schema
    `{"type": "object", "additionalProperties": {"type": "string", "format": "date"}}`
    would result into `{'key1': '2019-05-22'}` (master) instead of `{'key1': datetime.date(2019, 5, 22)}` (after this PR)
 * when default values are used they are not unmarshaled.
  Unmarshaling `{}` against the following schema
    `{"type": "object", "property": {"key1": {"type": "string", "format": "date", "default": "2019-05-22"}}}`
    would result into `{'key1': '2019-05-22'}` (master) instead of `{'key1': datetime.date(2019, 5, 22)}` (after this PR)
 * ~unmarshaling a polymorphic object into a dictionary (`unmarshal_object`) does not keep into account the identified polymorphic object and this misaligns `unmarshal_object` from `unmarshal_model`
  NOTE: respect #331 this PR still raises `SwaggerMappingError` in case of not found model (even if validation is disabled). This will be addressed on a separate PR~
 * unmarshaling a polymorphic object into a dictionary (`unmarshal_object`) does not keep into account the identified polymorphic object and this misaligns `unmarshal_object` from `unmarshal_model`
  NOTE: this PR removes the raise `SwaggerMappingError` in case of not found model as validation will take care of this (added test to ensure that this is the case)

### DIFF
--- a/bravado_core/unmarshal.py
+++ b/bravado_core/unmarshal.py
@@ -153,16 +153,9 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
     discriminator = object_spec.get('discriminator')
     if discriminator is not None:
         child_model_name = object_value.get(discriminator, None)
-        child_model_type = swagger_spec.definitions.get(child_model_name, _NOT_FOUND)
-        if child_model_type is _NOT_FOUND:
-            # TODO: should this be here? not think so
-            raise SwaggerMappingError(
-                'Unknown model {0} when trying to unmarshal {1}. '
-                'Value of discriminator {2} did not match any definitions.'
-                .format(child_model_name, object_value, discriminator)
-            )
-        elif child_model_type._model_spec != object_spec:
-            # The discriminator field does targets a different model, let's unmarshal it
+        child_model_type = swagger_spec.definitions.get(child_model_name, None)
+        if child_model_type and child_model_type._model_spec != object_spec:
+            # The discriminator field does targets a different model, let's unmarshal with it
             return unmarshal_schema_object(swagger_spec, child_model_type._model_spec, object_value)
 
     result = {}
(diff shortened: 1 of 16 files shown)


bravado_core/_compact.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
sjaensch: Why `_compact`? I'd prefer `_compat` / `_compatibility`

bravado_core/schema.py
```diff
@@ -108,16 +103,7 @@ def get_spec_for_prop(swagger_spec, object_spec, object_value, prop_name, proper
...
-        # If the de-referenced specification is for a x-nullable property
-        # then copy the spec and add the x-nullable property.
-        # If in the future there are other attributes on the property that
-        # modify a referenced schema, it can be done here (or rewrite
-        # unmarshal to pass the unreferenced property spec as another arg).
-        if 'x-nullable' in prop_spec and 'x-nullable' not in result_spec:
-            result_spec = copy.deepcopy(result_spec)
-            result_spec['x-nullable'] = prop_spec['x-nullable']
-        return result_spec
+        return deref(prop_spec)
```
sjaensch: While this is correct, it's behavior-changing, right? Any reason you included it in this PR?

bravado_core/util.py
```diff
@@ -70,8 +70,13 @@ def __get__(self, obj, cls):
         return value
 
 
+class RecursiveCallException(Exception):
+    pass
+
+
 def memoize_by_id(func):
     cache = func.cache = {}
+    on_fly_cache = set()
```
sjaensch: How about `recursion_detector` or `key_in_progress_set`?

bravado_core/unmarshal.py
```diff
@@ -131,6 +143,22 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
...
     properties = collapsed_properties(object_spec, swagger_spec)
 
+    # Check if object_spec is polymorphic
+    discriminator = object_spec.get('discriminator')
+    if discriminator is not None:
+        child_model_name = object_value.get(discriminator, None)
+        child_model_type = swagger_spec.definitions.get(child_model_name, _NOT_FOUND)
+        if child_model_type is _NOT_FOUND:
+            # TODO: should this be here? not think so
+            raise SwaggerMappingError(
```
sjaensch: What have we done in the past? I guess we have unmarshalled using the "wrong" object spec...? Should we keep this backwards-compatible and instead log something?
On the other hand this seems like a bug to me... can we instead do a swagger-spec-validator change first that would catch this?

bravado_core/schema.py
macisamuele: It was included due to coverage change.
The change was caused by the fix of `test_nullable_object_properties` done to reduce warnings.
As it is out of scope I silenced the warning with a different approach without modifying the test code (so continuing to check that the removed code is tested)

bravado_core/unmarshal.py
macisamuele: In the past we would just unmarshal according to the base schema.

I've add the exception (with the TODO) to preserve consistency between unmarshaling as dictionary or as model.

This will change will be addressed on a following PR where I do remove validation from the unmarshaling flow (as it should not be done in the first place ;) )

Is this approach ok to you?

macisamuele: Addressed in this CR as well

PR #336, opened 2019-05-23
Enhance unmarshaling performances

macisamuele:
**NOTE**: This CR replaces #331 . This is done due to the fact that history on the other PR is quite confusing.

The goal of this PR is, as mentioned in the title, unmarshaling performance enhancement.
NOTE: A similar PR will be published for applying the same approach described here for the marshaling process. I would like to have this verified first in order to avoid having two massive branches open at the same time

In order to achieve unmarshaling performance boost I've _rewritten_ the whole `bravado_core.unmarshal` module.

The current, before this PR, logic of unmarshaling is:
 1. extract the type from the schema
 2. if the schema if of type x then call the appropriate unmarshaling function
     * if the schema is of type object: call the top unmarshaling function for the property (point 1)
     * if the schema is of type object: call  the top unmarshaling function for each item of the list (point 1)
     * if the schema is of primitive type: identify, each time, the right formatting function to use
 3. once the unmarshaling function is identified then perform the real unmarshaling

This PR aims to cache results of point 1 and 2 for later "cost-free" usage.

While making this changes I've noticed that there might be no real good reason to have all the unmarshaling functions public, users of the library commonly use `bravado_core.unmarshal.unmarshal_schema_object`.
In order to simplify future maintenance work I've marked all the other functions as deprecated and will be removed in the next major release. Doing so will allows us, if needed, to modify more the internals of unmarshaling without worrying about breaking backward compatibility.

Let's get to the juicy part 😄 
Raw data on: https://gist.github.com/macisamuele/9ff23d3e5e81c40cd884425a7eae5751

This PR provides a massive performance improvement:
 * the worst performance improvement is ~46% (if in the branch takes 1 then on master takes 1.87)
 * the best performance improvement is ~70% (if in the branch takes 1 then on master takes 3.33)
 * the average performance improvement is ~60% (if the branch takes 1 then on master takes 2.57)

---

Be aware that:
 * `bravado_core.unmarshal.get_unmarshaling_method` is wrapped by `_decorators.wrap_recursive_call_exception` and `memoize_by_id` decorators.
    `_decorators.wrap_recursive_call_exception` decorator is needed as models could be recursive and so invoking `get_unmarshaling_method` might end up on an infinite recursion.
    In order to proactively deal with it I've modified `@memoize_by_id` decorator to raise a controlled exception if an unbounded recursion is identified (ℹ️ the interpreter would raise anyway with a `RuntimeError: maximum recursion depth exceeded`)
    `memoize_by_id` decorator is needed to speed up evaluation of the marshaling method for the same spec and schema. This effectively is one of the major performance boost (basically the expensive point 1 and 2 of the original flow become, more or less, a dictionary lookup)
 * All the test changes have been made before the refactoring (except `tests/unmarshal/unmarshal_object_test.py`) to guarantee that functionally nothing changed ;)


bravado_core/_decorators.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""
+The module defines useful decorators for the marshaling and unmarshaling process
+
+NOTE: The module has been marked as PRIVATE., so we are required to preserve
+      This is done to allow future modification to the defined decorators
```
sjaensch: Let's clean up this docstring. :)

bravado_core/model.py
```diff
@@ -366,6 +369,10 @@ def __init_from_dict(self, dct, include_missing_properties=True):
         for attr_name in additional:
             self.__dict[attr_name] = dct[attr_name]
 
+    @lazy_class_attribute
```
sjaensch: Is this worth caching? The cache lookup should have more or less the same cost as recalculating the value, right?

bravado_core/model.py
```diff
@@ -391,6 +398,20 @@ def _deny_additional_properties(self):
     def _include_missing_properties(self):
         return self._swagger_spec.config['include_missing_properties']
 
+    @lazy_class_attribute
```
sjaensch: Same question here about the value of caching.

bravado_core/model.py
```diff
@@ -581,8 +602,8 @@ def _unmarshal(cls, val):
         :type val: dict
         :rtype: .Model
         """
-        from bravado_core.unmarshal import unmarshal_model
-        return unmarshal_model(cls._swagger_spec, cls._model_spec, val)
+        from bravado_core.unmarshal import unmarshal_schema_object
+        return unmarshal_schema_object(cls._swagger_spec, cls._model_spec, val)
```
sjaensch: Why are you changing this?

tests/_decorators_test.py
```diff
@@ -0,0 +1,76 @@
...
+def test_handle_null_value(minimal_swagger_spec, object_schema, value, expected_value):
+    @handle_null_value(minimal_swagger_spec, object_schema)
+    def foo(param1):
+        return param1
+
+    if isinstance(expected_value, type) and issubclass(expected_value, Exception):
+        with pytest.raises(expected_value):
+            foo(value)
+    else:
+        assert foo(value) == expected_value
```
sjaensch: Please split this test up into two separate cases, one where an exception is raised on one where it isn't.

bravado_core/unmarshal.py
```diff
@@ -130,95 +92,239 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
...
+    discriminator_property,
+    model_to_unmarshaling_function_mapping,
+    model_type,
+    include_missing_properties,
+    properties_to_default_value,
+    additional_properties_unmarshaling_function,
+    model_value,
+):
+    """
+    Unmarshal a dict into a Model instance or a dictionary (according to the 'use_models' swagger_spec configuration).
```
sjaensch: It looks to me like you're potentially treating any `dict` as model, even the ones where we know they're _not_ a model. Conflating these use cases while keeping the term model is confusing - please separate them out or use a different naming scheme.

bravado_core/unmarshal.py
```diff
@@ -130,95 +92,239 @@ def unmarshal_object(swagger_spec, object_spec, object_value):
...
+def _unmarshal_model(
+    properties_to_unmarshaling_function,
+    discriminator_property,
+    model_to_unmarshaling_function_mapping,
+    model_type,
+    include_missing_properties,
+    properties_to_default_value,
+    additional_properties_unmarshaling_function,
+    model_value,
+):
```
sjaensch: This function could use a bit more documentation. For example, why do we need `model_to_unmarshaling_function_mapping`? It has to do with discriminated models, right? Why do we need it? Can't we look that up in here? And why do we need model-specific unmarshaling functions at all?

macisamuele: In order to avoid performance penalties while unmarshaling polymorphic schemas it would be better to already have the possible models definitions that the objected could be discriminated with.

We don't actually need to have the unmarshaling function already present, as obtaining it would be fast (if the function is alredy evaluated), so I'm basically reducing the amount of info that we're caching with a negligible performance different and more importantly I'm updating the parameter name as for sure the one that was defined was not as helpful as I did expect.

.coveragerc
```diff
@@ -1,7 +1,23 @@
...
+    .tox/*
+    venv/*
+    /usr/*
+    setup.py
 exclude_lines =
     def __repr__
     raise AssertionError
     raise NotImplementedError
+    if typing.TYPE_CHECKING:
+    if getattr\(typing, 'TYPE_CHECKING', False\):
```
sjaensch: We shouldn't be using `if typing.TYPE_CHECKING`, since that was added only in Python 3.5.2. So let's not ignore it for coverage but instead fix any remaining occurences. :)

bravado_core/_compat_typing.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+import typing
+from mypy_extensions import Arg
+from typing import NoReturn  # NOTE: This is available on Python3.5.4+  # noqa: F401
```
sjaensch: Shouldn't we handle it still? Doesn't it mean mypy runs will fail for target Python versions 3.5.2 and 3.5.3?

mypy.ini
```diff
@@ -0,0 +1,23 @@
+[mypy]
+ignore_missing_imports = True
+python_version = 2.7
+strict_optional = True
+warn_redundant_casts = True
+disallow_untyped_calls = False
+check_untyped_defs = False
```
sjaensch: Do you get too many errors if you enable this? Typically although we lack a lot of type information we should be able to tell mypy to check the code already.

setup.py
```diff
@@ -48,8 +35,22 @@
...
+        "python-dateutil",
+        "pyyaml",
+        "simplejson",
+        "six",
+        "swagger-spec-validator>=2.0.1",
+        "pytz",
+        "msgpack-python>=0.5.2",
+    ],
+    package_data={
+        'bravado': ['py.typed'],
```
sjaensch: bravado-core :)

bravado_core/_compat_typing.py
macisamuele: That's possible but I cannot really guarantee that.
I'm going for the easy path of using `-> None` instead.

setup.py
macisamuele: lol

mypy.ini
macisamuele: setting `check_untyped_defs = True` would cause reporting on multiple files
```
13:50 $ mypy bravado_core | cut -d: -f1 | sort | uniq -c
      4 bravado_core/docstring.py
      4 bravado_core/model.py
      1 bravado_core/operation.py
      1 bravado_core/param.py
      1 bravado_core/resource.py
      5 bravado_core/spec.py
      2 bravado_core/spec_flattening.py
      1 bravado_core/util.py
```
I'm opting for a solution that allows to ensure properly defined types in a subset of files.
PR #314 will take care of adding the typing where missing

.coveragerc
macisamuele: ack

bravado_core/_compat_typing.py
sjaensch: Yeah we could test it out. Maybe a possible solution would be (haven't tested it):

    try:
        from typing import NoReturn
    except ImportError:
        NoReturn = None

And we get the advantage of improved type checking if you are on a new enough Python version, plus the annotation is better for documentation purposes.

mypy.ini
sjaensch: yeah sounds good!

bravado_core/_compat_typing.py
macisamuele: Using the try except approach

PR #338, opened 2019-05-27
Use partials as they do not introduce the overhead of wrapping functions

macisamuele:
Double checking the performance improvements of #336, I've noticed that the final version (the one that has been merged) does have equivalent performances of the "original" code.

Comparing performance results between the individual commits I've found a specific commit responsible for the majority of the performance issues:  https://github.com/Yelp/bravado-core/commit/bfa6ceec0cf817a397841602a1a4b4281e8fd77e

In order to restore performances, this PR is going to re-introduce the usage of partials.
The drawback of partials is that we might not have accurate type checking on the arguments of the function, but it provides way better performances that a _normal_ function call.

Raw data on https://gist.github.com/macisamuele/785ae30d49ab94244734fb5697930893
The changes, introduced during the #336 did reduce the improvements but #337 should help as well

### DIFF
--- a/bravado_core/unmarshal.py
+++ b/bravado_core/unmarshal.py
@@ -177,7 +177,7 @@ def _get_unmarshaling_method(swagger_spec, object_schema, is_nullable=True):
         return _no_op_unmarshaling
     else:
         return partial(
-            _unknown_type_unmarhsaling,
+            _unknown_type_unmarshaling,
             object_type,
         )
 
@@ -187,7 +187,7 @@ def _no_op_unmarshaling(value):
     return value
 
 
-def _unknown_type_unmarhsaling(object_type, value):
+def _unknown_type_unmarshaling(object_type, value):
     # type: (typing.Union[typing.Type[dict], typing.Type[Model]], typing.Any) -> NoReturn
     raise SwaggerMappingError(
         "Don't know how to unmarshal value {0} with a type of {1}".format(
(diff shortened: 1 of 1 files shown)


bravado_core/unmarshal.py
```diff
@@ -236,15 +234,10 @@ def _unmarshaling_method_array(swagger_spec, object_schema):
...
-        # type: (typing.Any) -> typing.Any
-        return _unmarshal_array(
-            unmarshal_array_item_function=_get_unmarshaling_method(swagger_spec=swagger_spec, object_schema=item_schema),
-            value=value,
-        )
-
-    return wrapper_array
+    return partial(
+        _unmarshal_array,
+        _get_unmarshaling_method(swagger_spec, item_schema),
```
macisamuele:
```suggestion
        _get_unmarshaling_method(swagger_spec=swagger_spec, object_schema=item_schema),
```

bravado_core/unmarshal.py
```diff
@@ -176,12 +176,10 @@ def _get_unmarshaling_method(swagger_spec, object_schema, is_nullable=True):
...
         return _no_op_unmarshaling
     else:
-        @wraps(_unknown_type_unmarhsaling)
-        def wrapper_unknown(value):
-            # type: (typing.Any) -> typing.Any
-            return _unknown_type_unmarhsaling(object_type=object_type, value=value)
-
-        return wrapper_unknown
+        return partial(
+            _unknown_type_unmarhsaling,
```
sjaensch: typo in the function name :)

PR #339, opened 2019-05-27
Enhance marshaling performances

macisamuele:
This PR mimics what has been done in #336 for the unmarshaling process.
As for the unmarshaling also in case of the marshaling process we have a good performance improvements.

Raw data: https://gist.github.com/macisamuele/ee3b66176becab678aedf943e7a0e512
* the worst performance improvement is ~61% (if in the branch takes 1 then on master takes 2.58)
* the best performance improvement is ~72% (if in the branch takes 1 then on master takes 3.60)
* the average performance improvement is ~79% (if the branch takes 1 then on master takes 4.89)

### DIFF
--- a/bravado_core/_decorators.py
+++ b/bravado_core/_decorators.py
@@ -15,8 +15,8 @@ if getattr(typing, 'TYPE_CHECKING', False):
 
 
 @memoize_by_id
-def handle_null_value(swagger_spec, object_schema, is_nullable=False):
-    # type: (Spec, JSONDict, bool) -> typing.Callable[[FuncType], FuncType]
+def handle_null_value(swagger_spec, object_schema, is_nullable=False, is_marshaling_operation=False):
+    # type: (Spec, JSONDict, bool, bool) -> typing.Callable[[FuncType], FuncType]
     # TODO: remove is_nullable support once https://github.com/Yelp/bravado-core/issues/335 is addressed
     """
     Function wrapper that performs some check to the wrapped function parameter.
@@ -31,9 +31,6 @@ def handle_null_value(swagger_spec, object_schema, is_nullable=False):
      * nullable parameter is needed to ensure that x-nullable is propagated in the case it is defined
        as sibling in a reference object (ie. `{'x-nullable': True, '$ref': '#/definitions/something'}`)
 
-    :type swagger_spec: :class:`bravado_core.spec.Spec`
-    :type object_schema: dict
-    :type is_nullable: bool
     """
     default_value = schema.get_default(swagger_spec, object_schema)
     is_nullable = is_nullable or schema.is_prop_nullable(swagger_spec, object_schema)
@@ -53,7 +50,8 @@ def handle_null_value(swagger_spec, object_schema, is_nullable=False):
                         raise SwaggerMappingError(
                             'Spec {0} is a required value'.format(object_schema),
                         )
-
+                elif is_marshaling_operation:
+                    return value
             return func(value)
 
         return wrapper
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -253,7 +253,6 @@ def _marshaling_method_file(swagger_spec, object_schema):
 
 def _marshal_object(
     swagger_spec,  # type: Spec
-    model_type,  # type: typing.Union[typing.Type[JSONDict], typing.Type[Model]]
     properties_to_marshaling_function,  # type: typing.Dict[typing.Text, MarshalingMethod]
     additional_properties_marshaling_function,  # type: MarshalingMethod
     discriminator_property,  # type: typing.Optional[typing.Text]
@@ -267,7 +266,6 @@ def _marshal_object(
     Marshal a dict or Model instance into its JSON Object representation.
 
     :param swagger_spec: Spec object
-    :param model_type: Type of the return value (:class:`dict` or a subclass of :class:`bravado_core.model.Model`)
     :param properties_to_marshaling_function: Mapping between property name and associated unmarshaling method
     :param additional_properties_marshaling_function: Unmarshaling function of eventual additional properties
     :param discriminator_property: Discriminator property name. It will be `None` if the schema is not a polymorphic schema
@@ -280,8 +278,8 @@ def _marshal_object(
     """
     if not is_dict_like(model_value) and not isinstance(model_value, Model):
         raise SwaggerMappingError(
-            "Expected type to be {0} to marshal value '{1}' to a dict. Was {2} instead.".format(
-                'dict' if model_type is dict else 'dict or Model', model_value, type(model_value),
+            "Expected type to be dict or Model to marshal value '{0}' to a dict. Was {1} instead.".format(
+                model_value, type(model_value),
             ),
         )
 
@@ -375,7 +373,6 @@ def _marshaling_method_object(swagger_spec, object_schema):
     return partial(
         _marshal_object,
         swagger_spec,
-        model_type if model_type and swagger_spec.config['use_models'] else dict,
         properties_to_marshaling_function,
         additional_properties_marshaling_function,
         discriminator_property,
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -4,7 +4,6 @@ from functools import partial
 
 import typing
 from six import iteritems
-from six import iterkeys
 
 from bravado_core import _decorators
 from bravado_core import schema
@@ -252,7 +251,7 @@ def _marshaling_method_file(swagger_spec, object_schema):
     return _no_op_marshaling
 
 
-def _marshal_model(
+def _marshal_object(
     swagger_spec,  # type: Spec
     model_type,  # type: typing.Union[typing.Type[JSONDict], typing.Type[Model]]
     properties_to_marshaling_function,  # type: typing.Dict[typing.Text, MarshalingMethod]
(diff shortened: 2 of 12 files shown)


bravado_core/_decorators.py
```diff
@@ -15,8 +15,8 @@
 
 
 @memoize_by_id
-def handle_null_value(swagger_spec, object_schema, is_nullable=False):
-    # type: (Spec, JSONDict, bool) -> typing.Callable[[FuncType], FuncType]
+def handle_null_value(swagger_spec, object_schema, is_nullable=False, marshaling_operation=False):
```
sjaensch: Please use `is_marshaling_operation` as a name.

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
-            if not is_prop_nullable(swagger_spec, prop_spec):
-                continue
+def _no_op_marshaling(value):
+    # type: (typing.Any) -> typing.Any
+    return value
 
-        result[k] = marshal_schema_object(swagger_spec, prop_spec, v)
 
-    return result
+def _unknown_type_marhsaling(object_type, value):
```
sjaensch: marhsaling -> marshaling

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
     :rtype: dict
     :raises: SwaggerMappingError
     """
-    deref = swagger_spec.deref
+    warnings.warn(
+        'marshal_object will be deprecated in the next major release. '
+        'Please use the more general entry-point offered in marshal_schema_object',
+        DeprecationWarning,
+    )
+    null_decorator = _decorators.handle_null_value(swagger_spec=swagger_spec, object_schema=object_spec)
```
sjaensch: Why are you not setting `marshaling_operation` to `True` here? Same for `marshal_model`.

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
+    additional_properties_marshaling_function,  # type: MarshalingMethod
+    discriminator_property,  # type: typing.Optional[typing.Text]
+    possible_discriminated_type_name_to_model,  # type: typing.Dict[typing.Text, Model]
+    required_properties,  # type: typing.Set[typing.Text]
+    nullable_properties,  # type: typing.Set[typing.Text]
+    model_value,  # type: typing.Any
+):
+    # type: (...) -> typing.Any
+    """
+    marshal a dict into a Model instance or a dictionary (according to the 'use_models' swagger_spec configuration).
```
sjaensch: That's not what this function should be doing, it should be converting a model value (either a model instance or a dict) into the wire format, which would be a dict.

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
+
+    :type model_type: Model
+    :type model_value: dict
```
sjaensch: This is not correct, and not what your type annotation says. :)

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
+    )
+
+
+def _marshaling_method_file(swagger_spec, object_schema):
+    # type: (Spec, JSONDict) -> MarshalingMethod
+    # TODO: Type file is not a valid type. It is present to support parameter marshaling (move to bravado_core.param)  # noqa: E501
+    return _no_op_marshaling
+
+
+def _marshal_model(
```
sjaensch: This function is quite concerning to me. That is a _lot_ of new code for doing something that we already had. Why do we need so much new code in here and in `_marshaling_method_object`?

bravado_core/marshal.py
```diff
@@ -131,71 +108,285 @@ def marshal_object(swagger_spec, object_spec, object_value):
...
+        possible_discriminated_type_name_to_model,
+        required_properties,
+        nullable_properties,
+    )
 
-    if model_value is None:
-        return handle_null_value(swagger_spec, model_spec)
 
-    if not isinstance(model_value, Model):
+def _low_level_to_wire(swagger_spec, object_schema, swagger_format, value):
```
sjaensch: What does `low_level` mean here?

bravado_core/marshal.py
macisamuele: On it (this is copy and paste from unmarshaling methods and typo ;) )

bravado_core/_decorators.py
macisamuele: Fair point ;)

bravado_core/marshal.py
macisamuele: Aware of that ... docstrings are not-correct

bravado_core/marshal.py
macisamuele: Updating docstring

bravado_core/marshal.py
macisamuele: Because I forgot to fix them ... fixing as I type

bravado_core/marshal.py
macisamuele: The function is actually a wrapper around `SwaggerFormat.to_wire` to ensure that `SwaggerMappingError` is eventually raised instead of a generic exception.

Going to rename the function to `_marshal_primitive_type` to be more consistent with the equivalent methods for array and object

bravado_core/marshal.py
macisamuele: I did use the same approach used for unmarshaling, but I've not checked what is the time penalty that we might have if we use to "original" method.

I'll gather more data before continuing

macisamuele: Checked the approach and the added complexity was caused by the fact that `_marshal_model`  could receive models and or dictionaries.
In case a model is received we cannot iterate over the defined keys as we could do on a dictionary.
I'm working this around by iterating over the value keys (`Model` implements `__iter__`).

By doing so we can have a very simplified version of the method (that also is more similar to the original definition).

Side note:
 1. `_marshal_model`  will be renamed to `_marshal_object` to keep consistency with the unmarshaling flow
 2. I've tried implementing `iter` and `iteritems` in model and it lead to even nicer performances, but it might hide "attributes" (`Model.<property_name>`) that we were allowed to use

bravado_core/marshal.py
```diff
@@ -249,12 +264,18 @@ def _marshal_model(
...
     # type: (...) -> typing.Any
     """
-    marshal a dict into a Model instance or a dictionary (according to the 'use_models' swagger_spec configuration).
-
-    :type model_type: Model
-    :type model_value: dict
+    Marshal a dict or Model instance into its JSON Object representation.
+
+    :param swagger_spec: Spec object
+    :param model_type: Type of the return value (:class:`dict` or a subclass of :class:`bravado_core.model.Model`)
```
sjaensch: Where is this used? Is it only for the contruction of the SwaggerMappingError? Then we can probably find a better way to do this, or at least have a better interface here? Why do we need to pass in an actual `dict` class?

bravado_core/marshal.py
```diff
@@ -249,12 +264,18 @@ def _marshal_model(
...
+    :param properties_to_marshaling_function: Mapping between property name and associated unmarshaling method
+    :param additional_properties_marshaling_function: Unmarshaling function of eventual additional properties
+    :param discriminator_property: Discriminator property name. It will be `None` if the schema is not a polymorphic schema
+    :param possible_discriminated_type_name_to_model: Mapping of the possible dereferenced Model names and Model instances.
+    :param required_properties: Set of required properties of the object schema
+    :param nullable_properties: Set of nullable properties of the object schema
+    :param model_value: JSON value to marshal
```
sjaensch: This is not a JSON value but the dict or model that is to be marshaled.

bravado_core/_decorators.py
```diff
@@ -15,8 +15,8 @@
 
 
 @memoize_by_id
-def handle_null_value(swagger_spec, object_schema, is_nullable=False):
-    # type: (Spec, JSONDict, bool) -> typing.Callable[[FuncType], FuncType]
+def handle_null_value(swagger_spec, object_schema, is_nullable=False, is_marshaling_operation=False):
```
sjaensch: Just a general comment: Such behavior-changing flags are not a good idea for functions. You've already seen that with your own code, it's quite easy to forget them or not apply them correctly. It can be hard to detect that, since code just continues to work (most of the time).

In the future, I'd recommend creating two functions (one in marshal.py, one in unmarshal.py) that you then use in those files. They can both wrap a common function that maybe has that flag if there is no way around it. But at least there is then only two callsites to that common function, and all the other places will use the easier-to-use wrapper function.

macisamuele: I like this approach, I would like to have this done as follow-up to avoid increasing of complexity of this PR.
Does this work for you?

bravado_core/marshal.py
macisamuele: The "return value" is not accurate, going to fix that.
And yes, `model_type` is used only for the construction of `SwaggerMappingError`.
As the error message does not necessarily need to be selective on `dict` vs `dict or Model` I'm going to remove the parameter

bravado_core/marshal.py
macisamuele: Nice catch , checking the other cases I've noticed that there was the same "issue" on the array docstring as well

macisamuele:
@sjaensch I've updated the gist content and the description.
Performance improvements (according to the ``tests/profiling/marshal_profiler_test.py``) are bound between 61% and 79%

bravado_core/_decorators.py
macisamuele: Change is ready in https://github.com/Yelp/bravado-core/commit/e0fc28ea054c270a8fcfb13a1466360a1b3480d6 
As soon as this PR will be merged I'll post a new PR for the suggested improvement.

sjaensch: Yeah sure a follow-up sounds good to me. :)

PR #341, opened 2019-05-31
Update changelog

macisamuele:
This PR, as visible on the different commits, ensures that all Issues and PRs have a link to the related github resource. This is done to provide an easier access to the details of the changes listed on the changelog.

Additionally to the linking part:
 * I've fixed the usage of single back-quotes (use \`\`internally_dereference_refs\`\` instead of \`internally_dereference_refs\`) such that the code will be rendered as such.
 * Added the changelog for the new release. This is done mostly because there are multiple changes and it would be better to have a proper code-review for the wording.

⚠️: This branch should not be merged before #339 as the new version changes do refer to its changes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,17 @@ Changelog
 =========
 .. Make sure to link Issue and PR information as `(PR|Issue) #xxx`_ and with a link at the bottom of the document
 
+5.13.0 (2019-xx-xx)
+-------------------
+| The release is mainly based on performance improvements of marshalling and unmarshalling processes.
+
+- ``Model`` instances use ``__slots__`` and ``@lazy_class_attribute`` to reduce ``Model`` class footprint - `PR #330`_
+- ``unmarshal_schema_object`` honours the additional properties and default values - `PR #333`_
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -1,75 +1,87 @@
...
+5.13.0 (2019-xx-xx)
 -------------------
-- Prevent ``DeprecationWarning`` in Python 3.7 - PR #326
-- Ensure compatibility with Windows Platform - PR #327, #328
+| The release is mainly based on performance improvements of marshalling and unmarshalling processes.
+
+- ``Model`` instances use ``__slots__`` and ``@lazy_class_attribute`` to reduce ``Model`` class footprint - `PR #330`_
+- ``unmarshal_schema_object`` honours the additional properties and default values - `PR #333`_
+- ``unmarshal_model``/``unmarshal_object`` do not raise ``SwaggerMappingError`` in case of unknown polymorphic objects (better distinction between unmarshalling and validating processes) - `PR #333`_
+- Refactor ``bravado_core.unmarshal`` module to enhance runtime performances - `PR #336`_
```
sjaensch: performances -> performance (there is no plural when used like this, only when speaking about performing)

PR #343, opened 2019-05-31
Increase resilience of is_model and is_object methods

macisamuele:
While testing the library, I've played a bit importint the kubernetes swagger specs ([link](https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.15/api/openapi-spec/swagger.json)).
The specs were selected as a good example of very very big swagger specs.
During the testing I've uncovered an issue related to `is_models` and `is_object` methods.

The method do assume that `x-model` and `title` properties of the object are string, but this is not something that we could actually assume as the specs do not prevent the presence of the attribute as valid object properties.
For example
```yaml
definitions:
  model:
    properties:
      title:
        type: string
      x-model:
        type: string
```

The goal of this PR is to better handle the condition of:
 * `x-model` is not a `string`
 * `title` is not a `string`

In order to prevent future regression I extracted a small test-case that is able to trigger the issue.

The script using while testing (that helped me discovering the bug) is the one listed below
```python
import requests
spec_url = 'https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.15/api/openapi-spec/swagger.json'
Spec.from_dict(requests.get(spec_url).json())
```

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -8,6 +8,7 @@ from warnings import warn
 
 from six import add_metaclass
 from six import iteritems
+from six import string_types
 from swagger_spec_validator.ref_validators import attach_scope
 
 from bravado_core.schema import collapsed_properties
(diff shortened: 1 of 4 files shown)


macisamuele:
There are no specific performance differences as:
 * `is_object` is used during model discovery and to get the validation flow started.
    The most critical path is in the validation flow, but it is used only to have the flow started (so one extra check that is used once)
 * `is_model` is used only for doc-string generation, so no real performance hit

PR #345, opened 2019-06-03
Fix model isinstance regression

macisamuele:
The changes in https://github.com/Yelp/bravado-core/commit/9603483c348b9ba00d1288ba88aeef4a62c7e155 introduced `__isinstancecheck__` regression.

The following test would fail, while it should pass (as a `Dog` is not a `Cat`)
```python
def test_(polymorphic_spec):
    dog = polymorphic_spec.definitions['Dog']._unmarshal({'name': 'name', 'type': 'Dog', 'birth_date': '2017-11-02'})
    assert isinstance(dog, polymorphic_spec.definitions['Cat']) is False
```

The regression is caused by the fact that due to the fact that I've adopted slots `hasattr(self, '_model_spec')` would always be returned `True` so I updated it to `type(self) == ModelMeta`.
The interesting point is that `type(self)` is always going to be `ModelMeta` so there would be a very fast short circuit in the logic.
This highlights that we do not have a great coverage of "edge" cases on that but it could have been noticed by observing the coverage drop ([v5.12.1](https://coveralls.io/builds/23479354/source?filename=bravado_core/model.py#L226) vs [v5.13.0](https://coveralls.io/builds/23753966/source?filename=bravado_core/model.py#L242)).

In order to fix the regression we might use two alternative approaches:
 1. revert the breaking change -> this would remove the slots advantage
 2. update the `__instancecheck__` logic and define more accurate tests coverage

I opted for the second solution as it would allow us to save space while keeping an accurate determination of `isinstance`.


tests/model/model_test.py
```diff
@@ -263,3 +279,22 @@ def test_ensure_model_spec_does_not_contain_references_if_fully_dereference_is_e
...
+
+
+def test_isinstance_model_instance_is_recognized_properly(user_type, user_kwargs, cat_type):
+    assert isinstance(user_type(**user_kwargs), WrongModelClass) is False
+    assert isinstance(user_type(**user_kwargs), Model) is True
+    assert isinstance(user_type(**user_kwargs), user_type) is True
+    assert isinstance(user_type(**user_kwargs), cat_type) is False
+
+
+def test_isinstance_works_in_case_of_inheritance(polymorphic_spec):
```
sjaensch: Why did you add all of these `isinstance` tests? Is it using your new `__subclasscheck__` internally?

macisamuele: I wrote `isinstance` tests as I do expect that users verify if an instance is of a given type more often that if a class is subclassing an other class.
I might still duplicate them to double check `issubclass` as well.

`__subclasscheck__` is called internally while running `__instancecheck__` from a `ABCMeta` class -> https://github.com/python/cpython/blob/master/Lib/_py_abc.py#L92-L106

bravado_core/model.py
```diff
@@ -226,26 +226,27 @@ def _collect_models(container, json_reference, models, swagger_spec):
...
+        is_subclass = super(ModelMeta, cls).__subclasscheck__(subclass)
+        if not is_subclass:
+            try:
+                # Check whether cls is part any subclass._inherits_from
+                is_subclass = any(
+                    cls is value
+                    for key, value in iteritems(subclass._swagger_spec.definitions)
+                    if key in subclass._inherits_from
+                )
+                if subclass:
```
sjaensch: This should be `if is_subclass:` I assume? Which would mean that you have holes (or bugs?) in your test coverage as well. :)

tests/model/model_test.py
```diff
@@ -102,7 +118,7 @@ def test_model_is_instance_inherits_from(cat_swagger_spec, pet_type, pet_spec, c
     assert pet_type._isinstance(cat)
     assert isinstance(cat, cat_type)
     assert isinstance(cat, pet_type)
-    assert isinstance(cat, new_pet_type)
+    assert isinstance(cat, new_pet_type) is False
```
sjaensch: Didn't we have issues with this in the past? That we wanted to make sure two model classes for the same model are considered identical? Are we now ok for this not to be true anymore? I think there needs to be a bit more of an explanation as to why this is fine now.

bravado_core/model.py
macisamuele: Definitely correct, at the moment I'm trying to find bandwidth to double check the different possible cases and to make sure that we're not missing things out.

Mostly double checking that the fix is actually fixing the issue.

tests/model/model_test.py
macisamuele: I did remember now why we did that edge case.
I'll work this out to make sure that we're not disrupting it.

bravado_core/model.py
macisamuele: Tests are running `isinstance(...)` once per test, this means that we'll not notice the side-effect of tricking `cls._abc_negative_cache` and `cls._abc_cache`

So I would opt for a combination of bug and testing hole 😞

macisamuele: I've added `test_ensure_that_tricking_abc_attributes_do_not_alter_results` to specifically check those cases

tests/model/model_test.py
macisamuele: Added dedicated test case (`test_model_isinstance_model_class_generate_by_different_Spec_object`) to ensure that we're not mistakenly removing this

macisamuele:
I'm surprised by the AppVeyor failure considering that the equivalent build on my AppVeyor profile was successful: https://ci.appveyor.com/project/macisamuele/bravado-core/builds/25119125/job/jbc8hlx34xtudmia vs https://ci.appveyor.com/project/sjaensch/bravado-core/build/job/2uwq50fg9vc55pg1

I'll try to do some digging to understand if there might be some flakiness

PR #347, opened 2019-07-26
Provide more useful error message for response header validation.

brycedrennan:
Previously the header_name was not found in the error, which made it not very informative.

Example:
Before: `None is not of type 'string'`
After: `None is not of type 'string' for header 'X-Request-Id'`

### DIFF
--- a/bravado_core/response.py
+++ b/bravado_core/response.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import msgpack
+from jsonschema import ValidationError
 from six import iteritems
 
 from bravado_core.content_type import APP_JSON
@@ -239,6 +240,10 @@ def validate_response_headers(op, response_spec, response):
 
     for header_name, header_spec in iteritems(headers_spec):
         header_spec = deref(header_spec)
-        validate_schema_object(
-            op.swagger_spec, header_spec, response.headers.get(header_name),
-        )
+        try:
+            validate_schema_object(
+                op.swagger_spec, header_spec, response.headers.get(header_name),
+            )
+        except ValidationError as e:
+            e.message = "{0} for header '{1}'".format(e.message, header_name)
+            raise e
--- a/tests/response/validate_response_headers_test.py
+++ b/tests/response/validate_response_headers_test.py
@@ -62,3 +62,4 @@ def test_invalid_headers(op):
     with pytest.raises(ValidationError) as excinfo:
         validate_response_headers(op, response_spec, response)
     assert "is not of type 'integer'" in str(excinfo.value)
+    assert "X-Foo" in str(excinfo.value)


brycedrennan:
These failing tests were already failing in master but I fixed them in #348

bravado_core/response.py
```diff
@@ -239,6 +240,10 @@ def validate_response_headers(op, response_spec, response):
...
         header_spec = deref(header_spec)
-        validate_schema_object(
-            op.swagger_spec, header_spec, response.headers.get(header_name),
-        )
+        try:
+            validate_schema_object(
+                op.swagger_spec, header_spec, response.headers.get(header_name),
+            )
+        except ValidationError as e:
+            raise ValidationError("{0} for header '{1}'".format(e.message, header_name))
```
macisamuele: According to the definition of `ValidationError` ([here](https://github.com/Julian/jsonschema/blob/v3.0.1/jsonschema/exceptions.py#L131)) the exception does have information related to the validator that caused the issue (ie. type mismatch), path, etc. 

By raising a new `ValidationError` exception we're preserving the type of the raised exception but we're going to miss the additional contextual information provided by `jsonschema` library.

My recommendation here would be to catch the exception, on the caller side, to do something meaning-full with it and to modify the exception message in order to enhance the reported message (modify the `message` attribute of the `ValidationError` instance).


PS. As you have the operation object available in the method I would suggest to mention the `operationId` that caused the exception such that it will be even easier to spot the exact call that caused the issue (ie. `"{0} for header '{1}' of operation '{2}'".format(e.message, header_name, op.operation_id)`)

brycedrennan: I should have mentioned that I checked this. No contextual information is being lost.

You can see here that the Validation error was raised originally without additional information. https://github.com/Julian/jsonschema/blob/v3.0.1/jsonschema/_validators.py#L261

The caller can’t add the information about which header failed because that information isn’t available to it.

macisamuele: I see your point. Different versions, maybe future versions, of the `jsonschema` library might change this by adding the "contextual" information on the exception instance.
As it will be low-effort to preserve the eventual present information I would still suggest to enhance the exception instance instead of replacing it

brycedrennan: @macisamuele I've made the change you suggested.

bravado_core/response.py
```diff
@@ -239,6 +240,10 @@ def validate_response_headers(op, response_spec, response):
...
         header_spec = deref(header_spec)
-        validate_schema_object(
-            op.swagger_spec, header_spec, response.headers.get(header_name),
-        )
+        try:
+            validate_schema_object(
+                op.swagger_spec, header_spec, response.headers.get(header_name),
+            )
+        except ValidationError as e:
+            e.message = "{0} for header '{1}'".format(e.message, header_name)
```
macisamuele: Considering that the library does still offers python2 support it would be nice if we ensure unicode strings (just in case).

To do so please add `from __future__ import unicode_literals` or wrap the string (before format) by `six.text_type`

brycedrennan: None of the other strings in this file or the jsonschema file that the original message is generated from do this.  Making a change to start doing that it would seem better suited to a different PR and outside the scope of this one.

brycedrennan:
@macisamuele I've updated the test to ensure the header name is included in the error message.

bravado_core/response.py
macisamuele: The concern is that swagger specs might define an header like `ß-value` and `ß` is a unicode character.
On Python2.7 `"{0} for header '{1}'".format(e.message, header_name)` an exception will cause an exception
```
Python 2.7.16 (default, Apr  4 2019, 22:44:12)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from jsonschema import ValidationError
>>> import six
>>> e = ValidationError('some text')
>>> header_name = u'ß-value'
>>>
>>> "{0} for header '{1}'".format(e.message, header_name)
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xdf' in position 0: ordinal not in range(128)
>>>
>>> six.text_type("{0} for header '{1}'").format(e.message, header_name)
u"some text for header '\xdf-value'"
>>>
```

As it is possible that we're suffering of similar issues on the rest of the codebase I'm fine to postpone this to a follow-up CR (I can take care of doing it once this is merged)

brycedrennan: I believe that only ascii characters are valid for http headers: 
https://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.2.6.p.1
https://stackoverflow.com/questions/29688599/are-special-characters-in-http-header-names-valid

Issue #146, opened 2017-02-07
Review formats `byte` and `binary` for type `string` (OpenAPI 2.0)


lucafavatella:
Please refer to https://github.com/Yelp/bravado-core/pull/144#issuecomment-277810200

PR #351, opened 2019-10-18
Add option to automatically encode/decode strings with byte format

nickgaya:
Define a `SwaggerFormat` for the "byte" string format that  base64-encodes binary data to wire and base64-decodes from wire to a Python `bytes` object.

This behavior is not backward-compatible so I added a config option to enable it, disabled by default.

Related: #146


nickgaya:
@macisamuele, @lucafavatella: Any thoughts on this?

PR #354, opened 2019-10-22
Model class improvements

nickgaya:
* Document `_json_reference` attribute in `Model` docstring.
* Remove class attributes from `Model.__slots__`; also set `__slots__ = ()` on model subclasses
* Add type hints for `Model._swagger_spec`, `Model._model_spec`, and `Model._json_reference`.

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -6,6 +6,7 @@ import re
 from copy import deepcopy
 from warnings import warn
 
+import typing
 from six import add_metaclass
 from six import iteritems
 from six import string_types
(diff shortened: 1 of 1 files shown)


bravado_core/model.py
```diff
@@ -645,6 +656,7 @@ def create_model_type(swagger_spec, model_name, model_spec, bases=(Model,), json
 
     return type(
         str(model_name), bases, dict(
+            __slots__=(),  # More memory-efficient
```
sjaensch: While this should improve memory usage, doesn't it prevent assignment of new attributes after instance creation? This is a backwards-incompatible change, I'm pretty sure there's at least some users out there that do this. We'll have to think about whether this is a tradeoff we'd want to make.

nickgaya: Attribute assignment is already overridden by `Model.__setattr__()` so I don't think users will run into issues, unless they're trying to use `__weakref__` or access `__dict__` directly.  (Although the Python data model is pretty complicated so it's possible I'm missing something.)

sjaensch: Ah makes sense - so I guess this was just an oversight when adding `__slots__`. Let's try it out. :)

PR #356, opened 2019-10-25
Add type annotations to is_list_like, is_dict_like, is_ref

sjaensch:
`is_list_like` is called by bravado-asyncio, and is the last call it makes that is untyped. Typing it means we can enable `disallow_untyped_calls` in its mpy config.

I'm typing the other two is_* functions as well while I'm at it.


sjaensch:
Included in version 5.15.0.

PR #358, opened 2019-11-02
Collapse required properties from inherited models

nickgaya:
For models with an `allOf` schema property,  required attributes should be inherited from parent schemas.


sjaensch:
Included in version 5.15.0.

PR #360, opened 2019-11-23
Ensure that spec is deep copiable

macisamuele:
A `bravado_core.spec.Spec` instance is currently not deep copiable.
So running something like `deepcopy(Spec.from_dict(<valid_specs>))` results on an unbounded recursion error.

The goal of this PR is to address this issue by defining  the `__deepcopy__` dunder method.

### DIFF
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 from collections import defaultdict
+from copy import deepcopy
 
 from six import iteritems
 
@@ -95,6 +96,14 @@ class Resource(object):
         log.debug(u"Building resource '%s'", name)
         self.name = name
         self.operations = ops
+
+    def __deepcopy__(self, memo=None):
+        if memo is None:
+            memo = {}
+        return self.__class__(
+            name=deepcopy(self.name, memo=memo),
+            ops=deepcopy(self.operations, memo=memo),
+        )
 
     def __repr__(self):
         return u"%s(%s)" % (self.__class__.__name__, self.name)
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -125,8 +125,13 @@ class Resource(object):
         return self.operations.keys()
 
     def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+
+        if id(self) == id(other):
+            return True
+
         return (
-            isinstance(other, self.__class__) and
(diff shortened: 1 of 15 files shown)


tests/spec/Spec/build_test.py
```diff
@@ -293,3 +295,7 @@ def test_build_raises_in_case_of_duplicated_models_between_paths_and_definitions
     )
 
     assert expected_exception_string == str(exinfo.value)
+
+
+def test_spec_instance_is_deep_copaible(petstore_spec):
```
sjaensch: typo, it should be `test_spec_instance_is_deep_copyable`

The PR looks good to me, but test coverage is a bit weak. Your one-liner is using several `__eq__` methods that you're introducing in this PR to verify that the `__deepcopy__` methods work. Could you either add unit tests for the `__eq__` methods as well or (preferably) add a few more assertions here, which should not only be about equality, but also about making sure we really did a deep copy.

bravado_core/resource.py
```diff
@@ -125,8 +125,13 @@ def __dir__(self):
         return self.operations.keys()
 
     def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+
+        if id(self) == id(other):
+            return True
```
sjaensch: Wouldn't it make sense to put the id check first?

bravado_core/spec.py
```diff
@@ -137,31 +138,42 @@ def __init__(
...
+            return False
+
+        if id(self) == id(other):
+            return True
+
+        # If self and other are of the same type but not pointing to the same memory location then we're going to inspect
+        # all the attributes.
+        # NOTE: Few attributes have recursive references to Spec or do not define an equality method we're going to blacklist them
+        return all(
+            getattr(self, attr_name, None) == getattr(other, attr_name, None)
```
sjaensch: This seems wrong to me - it allows the equality check to pass if one instance has the attribute set, while the other hasn't.

macisamuele: I'm going to update the return statement in the following way:
 * going to use a for loop instead of using `all` as it can simplify reading
 * going to use `getattr(self, attr_name) != getattr(other, attr_name)` check
 * going to catch `AttributeError` exception to ensure that differences in the defined attributes cause equality check to return False

bravado_core/resource.py
macisamuele: 👍

PR #361, opened 2019-11-26
Enhance typing annotations

macisamuele:
This PR does extract part of the work done in #314.

Not all the models have yet typing so mypy has been configured to be strict on all the typed modules and more relaxed on the non-typed ones (bravado_core.docstring, bravado_core.model, bravado_core.operation, bravado_core.param, bravado_core.request, bravado_core.resource, bravado_core.response, bravado_core.spec, bravado_core.spec_flattening)

### DIFF
--- a/bravado_core/security_requirement.py
+++ b/bravado_core/security_requirement.py
@@ -41,7 +41,7 @@ class SecurityRequirement(object):
 
     @property
     def security_definitions(self):
-        # type: () -> typing.Mapping[typing.Text, SecurityDefinition]
+        # type: () -> typing.Dict[typing.Text, SecurityDefinition]
         return {
             security_name: self.swagger_spec.security_definitions[security_name]
             for security_name in six.iterkeys(self.security_requirement_spec)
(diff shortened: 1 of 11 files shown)


bravado_core/security_requirement.py
```diff
@@ -31,25 +41,29 @@ def __init__(self, swagger_spec, security_requirement_spec):
 
     @property
     def security_definitions(self):
-        return dict(
-            (security_name, self.swagger_spec.security_definitions[security_name])
+        # type: () -> typing.Mapping[typing.Text, SecurityDefinition]
```
sjaensch: the return type should be `typing.Dict` - always be as specific as possible for return types.

bravado_core/spec.py
```diff
@@ -273,6 +278,7 @@ def get_ref_handlers(self):
         return build_http_handlers(self.http_client)
 
     def _force_deref(self, ref_dict):
+        # type: (typing.Any) -> typing.Any
```
sjaensch: While this is technically correct it might drastically decrease the value of type annotations. Some investigation is needed, but we do a lot of derefing when marshaling/unmarshaling/validating, possibly erasing type information in the process. I wonder if it would make more sense to say that we return the same type as the type of the `is_ref` argument.

PR #362, opened 2019-11-29
Fix deepcopy issue on Python3.5.0

macisamuele:
While working on https://github.com/Yelp/bravado/pull/445, I've noticed that the deepcopy support added in #360 seems to not be working on Python3.5.0.

The goal of this PR is to fix the issue.

**NOTE**: This PR is still in draft as the issue is not solved, but as least we can  reproduce it via travis

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Changelog
 -------------------
 - Add equality and deepcopy support for Spec objects - `PR #360`_
 
+**WARNING**: Deep Copy might not work properly on CPython 3.5.0. More details on `PR #362`_ and `Issue25447 <https://bugs.python.org/issue25447>`_
+
 5.15.0 (2019-11-15)
 -------------------
 - Properly build the list of required properties for a schema with ``allOf`` from the sub-schemas - `PR #358`_ - Thanks nickgaya for your contribution!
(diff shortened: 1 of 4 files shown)


macisamuele:
I've dug a bit more into the test failures in Python 3.5.0 and I did found out that the issue is a known issue for the CPython implementation ([issue link](https://bugs.python.org/issue25447)) and that it has already been resolved.

Unfortunately, the `RefResolver` of jsonschema library does use lru_cache into the class attributes and so this triggers the issue.

I've not found an easy/nice way to work this around but at the same time I do not expect that a lot of our users are using Python 3.5.0 (which should be the only affected version

My proposal at this point would be to update the bravado-core changelog highlighting the fact that deepcopy might not work properly on CPython 3.5.0.

Test run on all versions is available at: https://gist.github.com/macisamuele/edbdfee3bfdd10c33cc40ab654d9cb71

CHANGELOG.rst
```diff
@@ -6,6 +6,8 @@ Changelog
 -------------------
 - Add equality and deepcopy support for Spec objects - `PR #360`_
 
+**WARNING**: Deep Copy might not work properly on CPython 3.5.0. More details on `PR #362`_ and `Issue25447 <https://bugs.python.org/issue25447>`_
```
sjaensch: I do not think this is a good pattern. If some part of our software is known to not work with an older Python version then either we need a workaround, disable that feature or discontinue support for that Python version.

I vote for the latter. There are three major Python releases since 3.5 out now, plus we don't need to drop support for all of Python 3.5, just 3.5`.0`. Other libraries do this as well, e.g. [aiohttp](https://aiohttp.readthedocs.io/en/stable/) requires Python 3.5.3+.

macisamuele: Agreed.
Going to drop this PR in favour of #366

PR #363, opened 2020-01-04
Ensure hashability of classes with custom equality check

macisamuele:
Ensure hashability of classes with custom equality check.

The issue has been noticed because [Yelp/swagger-spec-compatibility](https://github.com/Yelp/swagger-spec-compatibility) tests are broken (due to the fact that `bravado_core.spec.Spec` instances are not hashable).

In #360 equality methods have been added in order to increase testing coverage of deepcopy of instances.

Objects inheriting from `object` (aka all) have:
 * `__eq__` implemented as always returning `False`
 * `__hash__` implemented as returning a unique identifier of the specific instance (aka result of `id`)

Unfortunately, on Python, if you override one `__eq__` method then the "default" `__hash__` implementation is not is use, and so to make the object hashable again you need to implement `__hash__` yourself (what this CR does). ([Documentation reference](https://docs.python.org/3/reference/datamodel.html#object.__hash__))

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -465,6 +465,9 @@ class Model(object):
         else:
             del self.__dict[property_name]
 
+    def __hash__(self):
+        return id(self)
+
     def __eq__(self, other):
         """Check for equality with another instance.
 
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -58,6 +58,9 @@ class Operation(object):
         # (key, value) = (param name, Param)
         self.params = {}
 
+    def __hash__(self):
+        return id(self)
+
     def __eq__(self, other):
         if id(self) == id(other):
             return True
--- a/bravado_core/resource.py
+++ b/bravado_core/resource.py
@@ -124,6 +124,9 @@ class Resource(object):
         """
         return self.operations.keys()
 
+    def __hash__(self):
+        return id(self)
+
     def __eq__(self, other):
         if id(self) == id(other):
             return True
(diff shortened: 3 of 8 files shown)


bravado_core/model.py
```diff
@@ -465,6 +465,9 @@ def __delitem__(self, property_name):
         else:
             del self.__dict[property_name]
 
+    def __hash__(self):
+        return id(self)
```
sjaensch: While this seems like a quick fix, I think it violates one of the [assumptions of hashes](https://hynek.me/articles/hashes-and-equality/) and equality of objects.

If obj1 == obj2 then hash(obj1) == hash(obj2)

What you're doing here is the default for the `__hash__` method. Python 3 intentionally removes it since you implemented `__eq__`. I'm honestly not sure if this is what we want to do, but I also don't have a better idea.

macisamuele: You're right ... for some reason I was remembering the property in the opposite direction.
I'll be fixing it before merging.
The fix might be based on top of equality check, not great as it will depend on the size of the Swagger Specs but I don't expect this to be a big issue.

If we do this that this is an issue we might just want to remove the "custom" equality checks as they were added mostly for testing purposes

macisamuele: @sjaensch I'm planning to drop this PR and to create a new PR where the custom `__eq__` methods (that made `bravado_core.spec.Spec` not hashable) are replaced by a different method `is_equal`.

Technically that change will be backward incompatible, so I'd like to have your thoughts around it.

PR #364, opened 2020-02-07
Replace __eq__ with is_equal to allow hashability

macisamuele:
This PR deprecates #363 
As users of the library might be using hashability of `bravado_core.spec.Spec` and we don't have a strong need to have equality checks (other than for testing) I'm proposing to preserve the default equality checks and move the customly defined ones (PR #360) into a different method `is_equal` such that we can have an entry point of equality checks without altering the hashability properties of the objects.

NOTE: In #363 I was trying to define an `__hash__` method were needed.
Unfortunately defining hash functions is not a trivial task as it might be computationally expensive and grows with the size of the specs (which is usually when you'd like to use it even more).

---

From #363

> The issue has been noticed because [Yelp/swagger-spec-compatibility](https://github.com/Yelp/swagger-spec-compatibility) tests are broken (due to the fact that `bravado_core.spec.Spec` instances are not hashable).
> 
> In #360 equality methods have been added in order to increase testing coverage of deepcopy of instances.
> 
> Objects inheriting from `object` (aka all) have:
>  * `__eq__` implemented as always returning `False`
>  * `__hash__` implemented as returning a unique identifier of the specific instance (aka result of `id`)
> 
> Unfortunately, on Python, if you override one `__eq__` method then the "default" `__hash__` implementation is not is use, and so to make the object hashable again you need to implement `__hash__` yourself. ([Documentation reference](https://docs.python.org/3/reference/datamodel.html#object.__hash__))

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -58,7 +58,9 @@ class Operation(object):
         # (key, value) = (param name, Param)
         self.params = {}
 
-    def __eq__(self, other):
+    def is_equal(self, other):
+        # Not implemented as __eq__ otherwise we would need to implement __hash__ to preserve
+        # hashability of the class and it would not necessarily be performance effective
         if id(self) == id(other):
             return True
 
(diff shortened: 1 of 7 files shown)


macisamuele:
Test failures are addressed in https://github.com/Yelp/bravado-core/pull/365

PR #372, opened 2020-02-24
Pickling support

### LINKED ISSUE #370: Add pickle support to bravado_core.spec.Spec
Currently `Spec` is not pickable
```
>>> import pickle
>>> from bravado_core.spec import Spec
>>> spec_dict = {
...   "swagger": "2.0",
...   "info": {
...     "title": "Test",
...     "version": "1.0"
...   },
...   "paths": {},
...   "definitions": {}
... }
>>> pickle.dumps(Spec.from_dict(spec_dict))
  File "<stdin>", line 1, in <module>
AttributeError: Can't pickle local object 'build_http_handlers.<locals>.download'
>>>
```

This prevents libraries to use `Spec` instances across processes or just to pickle clients.

macisamuele:
The goal of this PR is to fix #370 

In order to make `bravado_core.spec.Spec` pickable I had to define `__getstate__` and `__setstate__` methods into `Spec` and `Resource` class.

The reason of this is that pickling an object usually gets all the object attributes and tries to pickle them:
 * `Resource` does define `__dir__` method which confuses the pickling process
 * `Spec` needs to deal with the fact that runtime created types (models) are not picklable. In order to make them pickable I'm pickling a representation of the arguments needed to rebuild the type.
Unfortunately this approach has the defect that restoring (unpickling) the object might depend on the version of bravdo_core in use (more importantly on the signature of `create_model_type`.
I think that this is eventually acceptable considering that the signature did not changed over time and that in case that would happen we could still ensure that `bavado_core.__version__` is encoded into the picklable object.

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -925,7 +925,7 @@ def model_discovery(swagger_spec):
         swagger_spec.definitions = tmp_spec.definitions
 
 
-def to_pickleable_representation(model_name, model_type):
+def _to_pickleable_representation(model_name, model_type):
     # type: (typing.Text, typing.Type[Model]) -> typing.Dict[typing.Text, typing.Any]
     """
     Extract a pickleable representation of the input Model type.
@@ -933,7 +933,7 @@ def to_pickleable_representation(model_name, model_type):
     Model types are runtime created types and so they are not pickleable.
     In order to workaround this limitation we extract a representation,
     which is pickleable such that we can re-create the input Model type
-    (via ``from_pickleable_representation``).
+    (via ``_from_pickleable_representation``).
 
     NOTE:   This API should not be considered a public API and is meant
             only to be used by bravado_core.spec.Spec.__getstate__ .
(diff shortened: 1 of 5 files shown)


macisamuele:
@sjaensch for a couple of weeks I'll be far from a keyboard in order to make progress on this.

Definitely the point that you're raising is correct and that's the reason why I won't push more on introducing a similar feature until we don't have a clear idea on how to handle it.

I'll let you know when I get enough time to break the implementation and try to identify more weak points.

bravado_core/model.py
```diff
@@ -923,3 +923,37 @@ def model_discovery(swagger_spec):
...
+    """
+    Extract a pickable representation of the input Model type.
+
+    Model types are runtime created types and so they are not pickable.
+    In order to workaround this limitation we extract a representation,
+    which is pickable such that we can re-create the input Model type
+    (via ``from_pickable_representation``).
+
+    NOTE:   This API should not be considered a public API and is meant
+            only to be used by bravado_core.spec.Spec.__getstate__ .
```
sjaensch, line 939: Let's use a leading underscore then for both method names, that should make it clearer.

bravado_core/model.py
```diff
@@ -923,3 +923,37 @@ def model_discovery(swagger_spec):
         # this ensures that the generated models have no references
         _run_post_processing(tmp_spec)
         swagger_spec.definitions = tmp_spec.definitions
+
+
+def to_pickable_representation(model_name, model_type):
```
sjaensch, line 928: It's _pickleable_ - able to be [pickled](https://en.wikipedia.org/wiki/Pickling). :)

macisamuele: Nice catch.

sjaensch:
Looks good to me, but I see mypy and a test failure in the travis build. Could you take a look?

macisamuele:
The failures are suspicious to me. I've run tests locally and they were green.

The only difference spotted between local environment and travis is the version of Python 3.7 in use (travis uses 3.7.1 while 3.7.7 is installed on my laptop).

I'll try to install 3.7.1 to ease the local debugging

sjaensch:
This has been included in bravado-core 5.17.0.

PR #373, opened 2020-02-24
Typing annotations: resource and operation

macisamuele:
The goal of this PR is to fully type the bravado_core.operation and bravado_core.resource modules.

In order to achieve proper typing I had to update the implementation of the `__repr__` methods as they were returning unicode strings while `object.__repr__` expect strings (this applies only to python 2).

I thought that adding the condition over the python version would be the best way of dealing with the issue instead of dropping an effective (but unsightly) `type: ignore`

### DIFF
--- a/bravado_core/operation.py
+++ b/bravado_core/operation.py
@@ -3,6 +3,7 @@ import logging
 
 import typing
 from six import iteritems
+from six import PY2
 
 from bravado_core.exception import SwaggerSchemaError
 from bravado_core.param import Param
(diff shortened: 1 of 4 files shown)


bravado_core/operation.py
```diff
@@ -175,10 +192,16 @@ def operation_id(self):
         return _sanitize_operation_id(op_id, self.http_method, self.path_name)
 
     def __repr__(self):
-        return u"%s(%s)" % (self.__class__.__name__, self.operation_id)
+        # type: () -> str
+        repr = u"{self.__class__.__name__}({self.operation_id})".format(self=self)
+        if PY2:
+            return repr.encode('ascii', 'backslashreplace')
```
sjaensch: I'm wondering if this could break something for users with Python 2. But I guess it's very unlikely.

PR #374, opened 2020-03-05
add DEFAULT_FORMATS['uuid'] in formatter.py

kedrrr:


bravado_core/formatter.py
```diff
@@ -200,4 +201,11 @@ def wrapper(validatable_primitive):
         validate=NO_OP,  # jsonschema validates integer
         description='Converts [wire]integer:int64 <=> python long',
     ),
+    'uuid': SwaggerFormat(
+        format='byte',
+        to_wire=lambda u: str(u) if isinstance(u, uuid.UUID) else u,
+        to_python=lambda s: s if isinstance(s, uuid.UUID) else uuid.UUID(s),
+        validate=NO_OP,  # jsonschema validates uuid
```
sjaensch: I don't think JSONSchema does validate UUID, does it?
I'm also hesitant to add a default format to bravado-core that is not defined in the OpenAPI specification. If we do add it, could you please mention it in the documentation as well?

PR #375, opened 2020-03-09
Tentative to remove overhead of models

macisamuele:
The goal of this PR is to reduce the performance gap present while unmarshaling schemas if `use_models` is enabled or disabled.

In order to achieve the performance parity between models disabled (so raw `dict`) and models enabled (so `bravado_core.model.Model`) I've ensured that `Model` class extends `dict`.

In order to prevent unexpected regressions I've made sure that no tests would have been modified after the changing `Model`  base class.
This is true except for the last commit (6374839) where there is a _backward_ incompatible change (from commit message)
> Remove check for additional properties while creating model
> This is needed to reduce even more the difference in Model vs dict creation.
> NOTE: Once a model is created there are no checks in case additional properties would have been added.

Some hightligths:
 * pre-PR there was ~50% performance penalty on using models
 * after-PR there is <1% performance penalty on using models

Raw data are available on https://gist.github.com/macisamuele/dcd58fafd470c80902643760e4eb87ea

### DIFF
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -335,17 +335,6 @@ class Model(dict):
     _model_spec = None  # type: typing.ClassVar[JSONDict]
     _json_reference = None  # type: typing.ClassVar[str]
 
-    def __init__(self, *args, **kwargs):
-        dict.__init__(self, *args, **kwargs)
-        if self._model_spec.get('additionalProperties') is False:
-            additional_properties = self._additional_props
-            if additional_properties:
-                raise AttributeError(
-                    "Model {0} does not have attributes for: {1}".format(
-                        type(self), list(additional_properties),
-                    ),
-                )
-
     @lazy_class_attribute
     def _properties(cls):
         return collapsed_properties(cls._model_spec, cls._swagger_spec)
--- a/bravado_core/model.py
+++ b/bravado_core/model.py
@@ -3,12 +3,13 @@ import abc
 import functools
 import logging
 import re
-from copy import deepcopy
+from itertools import chain
 from warnings import warn
 
 import typing
 from six import add_metaclass
 from six import iteritems
+from six import iterkeys
 from six import string_types
 from swagger_spec_validator.ref_validators import attach_scope
 
@@ -271,8 +272,11 @@ class ModelMeta(abc.ABCMeta):
         return is_subclass
 
 
+_NOT_FOUND = object()
+
+
 @add_metaclass(ModelMeta)
-class Model(object):
+class Model(dict):
     """Base class for Swagger models.
 
     Attribute access:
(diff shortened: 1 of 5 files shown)


tests/model/model_constructor_test.py
```diff
@@ -54,16 +52,6 @@ def test_additionalProperties_true(definitions_spec, user_type, user_kwargs):
     assert set(user) == set(definitions_spec['User']['properties'].keys()).union({'foo'})
 
 
-def test_additionalProperties_false(user_type, user_kwargs):
```
sjaensch: Why did you remove this test?

bravado_core/model.py
```diff
@@ -474,36 +400,44 @@ def __eq__(self, other):
...
+        else:
+            return dict.__eq__(self, other)
 
     def __dir__(self):
         """Return only property names (including additional)."""
-        return sorted(self.__dict.keys())
+        return sorted(iter(self))
+
+    def __iter__(self):
+        return iter(set(chain(iterkeys(self.__class__._properties), iterkeys(self))))
```
sjaensch: I don't think I fully understand how `self.__class__._properties` and the dict iterator relate to each other. In `_as_dict` you have an explicit check where you iterate over yourself and skip keys that are in `_properties`. Do you need to do that here as well?

bravado_core/unmarshal.py
```diff
@@ -293,15 +293,18 @@ def _unmarshal_object(
...
             property_name, additional_properties_unmarshaling_function,
         )
         unmarshaled_value[property_name] = unmarshaling_function(property_value)
+        unmarshaled_attributes.add(property_name)
 
     if swagger_spec.config['include_missing_properties']:
         for property_name, unmarshaling_function in iteritems(properties_to_unmarshaling_function):
-            if property_name not in unmarshaled_value:
+            # if property_name not in unmarshaled_value:
+            if property_name not in unmarshaled_attributes:
```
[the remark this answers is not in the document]
macisamuele: `unmarshaled_value` can be:
 * a `dict` instance and so checking the key presence on a set does not have specific advantages other than wasting some memory :(
 * a `Model` instance and in this case the check would require, in the worst case, 2 dictionary checks (`dict.__contains__(self, item) or item in self.__class__._properties` ← synthesised from the `Model.__contains__` method)

In order to keep consistent timings while using models or raw dictionaries I've added the set

PS: Definitely the comment needs to be removed (I was benchmarking the two approaches to evaluate differences) and considering the PR comment I think might be worth investing few bytes adding a descriptive comment in the code 😁

tests/model/model_constructor_test.py
macisamuele: From the PR descrption
> This is true except for the last commit (6374839) where there is a backward incompatible change (from commit message)
> 
> > Remove check for additional properties while creating model
> > This is needed to reduce even more the difference in Model vs dict creation.
> > NOTE: Once a model is created there are no checks in case additional properties would have been added.

Essentially before this PR was not possible to create a model defining additional properties if the model spec does specify that no additional properties are accepted.
model spec
```yaml
x-model: Model
type: object
properties:
  prop:
    type: integer
additionalProperties: false
```
On master the following python code would raise an exception
```python
m = Model(prop=1, additional=1)
```
but the following code, attempting to do the same change would work
```python
m = Model(prop=1)
m.additional=1
```
I do think that this is an "inconsistent" behaviour and removing this "safety" allows for an marginal performance improvement.

bravado_core/model.py
macisamuele: I have to admit, this is far from _nice_ code.

The idea was to allow iteration over all the properties defined in the Model instance (either the properties specified on the specifications and the additional properties eventually embedded)

Spec example
```yaml
x-model: Model
type: object
properties:
  prop:
    type: integer
```
Python code
```python
assert set(iter(Model(prop=1))) == {"prop"}
assert set(iter(Model())) == {"prop"}  # prop is present because part of the properties
assert set(iter(Model(prop=1, additional=1))) == {"prop", "additional"}
```

bravado_core/model.py
```diff
@@ -516,28 +450,40 @@ def _as_dict(self, additional_properties=True, recursive=True):
...
+        dct = {
+            prop_name: _get_property(prop_name)
+            for prop_name in self.__class__._properties
+        }
 
-                attr_val = new_attr_val if is_list else new_attr_val[0]
+        if additional_properties:
+            for prop_name, attr_val in iteritems(self):
+                if prop_name in self.__class__._properties:
+                    continue
```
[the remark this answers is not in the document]
macisamuele: ++

macisamuele:
@sjaensch hope that you're doing well. Sorry for the long turnaround for this PR.

I've tried to resume the work on this PR, I've addressed few comments.
In the next days I'll try to get back to speed with this as by running few quick benchmarks (just to revalidate the status-quo) I've noticed some small inconsistencies with the results (but I'd like to investigate a bit more before raising additional concerns)

PR #377, opened 2020-06-14
Accelerate YAML parsing when LibYAML is available

hiromu:
When a Swagger definition is provided in YAML, the current implementation uses a Python loader for parsing.
However, it can be accelerated by using a native loader when LibYAML is available.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,12 @@ deps =
 commands =
     python -m pytest --cov --capture=no --benchmark-skip {posargs:tests}
 
+[testenv:py37]
+commands =
+    # cover the case that native libyaml is not avaiable
+    pip install --force-reinstall pyyaml --global-option="--without-libyaml"
+    {[testenv]commands}
+
 [testenv:benchmark]
 basepython = python3.7
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -11,12 +11,6 @@ deps =
 commands =
     python -m pytest --cov --capture=no --benchmark-skip {posargs:tests}
 
-[testenv:py37]
-commands =
-    # cover the case that native libyaml is not avaiable
-    pip install --force-reinstall pyyaml --global-option="--without-libyaml"
-    {[testenv]commands}
-
 [testenv:benchmark]
 basepython = python3.7
 deps =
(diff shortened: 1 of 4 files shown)


hiromu:
Thank you for your reply. I updated this pull request to pass all the tests.

For the mypy issue, I simply skipped the line in the same manner as people discussing in the mypy repository: https://github.com/python/mypy/issues/1153

tox.ini
```diff
@@ -11,6 +11,12 @@ deps =
 commands =
     python -m pytest --cov --capture=no --benchmark-skip {posargs:tests}
 
+[testenv:py37]
+commands =
+    # cover the case that native libyaml is not avaiable
```
sjaensch, line 16: I don't think we should modify the standard `py37` test environment for that case. Either we create a separate one or we just don't test this case.

sjaensch: Feel free to just remove your changes to this file, and I will merge your branch.

hiromu: I had trouble keeping the coverage check and figured out that we should modify test environments to cover the case that native libyaml is not available.
Thus, as you suggested, I will separate the case as a new environment.

hiromu:
After hours of investigation, I found there are several issues in pip and pyyaml which we cannot easily address:

- Building pyyaml from its source with Python2.7 shows incompatibility in the environment on Travis due to the version of setuptools.
- The current version of pip does not support specifying both `--global-option` and `python_version` simultaneously.

Thus, I decided to explicitly skip building pyyaml in the case of Python 2.7. I believe this is a reasonable way under these restrictions.

tox.ini
sjaensch: I wouldn't worry too much about the coverage check, it's ok for it to go down a bit with changes like these. Looking at the latest changes, I feel you're trying to do too much to test what essentially is a documented feature of PyYAML. I really think we can go ahead with just the changes to `bravado_core/spec.py`.

hiromu: As you won't mind the degradation of the coverage rate, I removed the newly added test condition. I think it now ready to be merged.

sjaensch:
Thanks for your contribution @hiromu!

PR #379, opened 2020-09-30
Restrict pytest to <4.7 to fix build, add py36 and mypy to tox envlist

sjaensch:
This fixes the build in the same way I fixed it for bravado. It also makes sure we run mypy as well as tests on Python 3.6 locally, and fixes installation / usage of pre-commit on systems where `python2.7` is not in `/usr/bin`.


analogue (approved):
🚢

PR #381, opened 2021-03-25
Update Spec to return new resolver instances for validators fixing an IndexError in RefResolver due to it not being thread-safe

marcelruhf:
This resolves an issue which occasionally results in an `IndexError` due to a race condition.
See MLOCPAR-6374 in JIRA for context.

### DIFF
--- a/bravado_core/spec.py
+++ b/bravado_core/spec.py
@@ -161,6 +161,15 @@ class Spec(object):
             handlers=self.get_ref_handlers(),
         )
 
+    @property
+    def validation_resolver(self):
+        # type: () -> RefResolver
+        return RefResolver(
+            base_uri=self.origin_url or '',
+            referrer=self.spec_dict,
+            handlers=self.get_ref_handlers(),
+        )
+
     def is_equal(self, other):
         # type: (typing.Any) -> bool
         """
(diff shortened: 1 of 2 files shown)


bravado_core/spec.py
```diff
@@ -161,6 +161,15 @@ def resolver(self):
...
         )
 
+    @property
+    def validation_resolver(self):
+        # type: () -> RefResolver
+        return RefResolver(
+            base_uri=self.origin_url or '',
+            referrer=self.spec_dict,
+            handlers=self.get_ref_handlers(),
+        )
```
macisamuele, line 171: I'm not sure about what race-condtions are you referring to and neither what is the effect (I'm no longer a Yelp employee and as such I don't have JIRA access).

I would highly recommend to share information and minimal-reproducible cases.
This is meant to provide visibility to library users about the issues that are present, when they are addressed and/or what workarounds are known to mitigate the issue.

Still, are you able to reliably reproduce the issue? can this be automated? having such test would allow to not regress in the future
---

About the change, it seems an harmless change which would probably help you (I can't really tell as I don't know your problem) but I would be very sceptical around this specific change mostly because it would defeat multiple performance related tasks by recreating every time a brand new validator.
I would encourage to validate that no substantial performance regressions are introduced by this.

One last remark: I don't see CI running for this PR. This is probably caused by having github workflows not enabled for pull-requests https://github.com/Yelp/bravado-core/blob/382db874b7b838dcfd169b0ce490d6a447ad6ff2/.github/workflows/ci.yml#L3

cc: @analogue  as author of the github actions integration

PR #385, opened 2021-10-19
Pin to jsonschema<4 and update config for testing on recent Python versions

terencehonles:
This combines #383 and #384 with some extra cleanup to get github actions working properly (They seem to be disabled here, but I have them now passing on my fork https://github.com/terencehonles/bravado-core/actions/runs/1358474617)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,8 +7,10 @@ import pytest
 import simplejson as json
 import yaml
 from mock import Mock
+from six import binary_type
 from six import iteritems
 from six import iterkeys
+from six import text_type
 from six.moves.urllib import parse as urlparse
 from six.moves.urllib.request import pathname2url
 from six.moves.urllib.request import url2pathname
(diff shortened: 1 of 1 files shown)


terencehonles:
From https://github.com/Yelp/bravado-core/pull/383#issuecomment-946440243

> I have https://github.com/terencehonles/bravado-core/tree/fix-deprecation-warning-for-RefResolver.in_scope which fixes the deprecation warning, but when testing the change I noticed there was an error during resolving. I checked and this looks like it only affects `Spec.from_dict` (I skipped all tests that go down that code path and the suite passed). It doesn't appear to be straightforward to find out what changed since I'm not familiar with this package or jsonschema but it looks like it's related to https://github.com/Julian/jsonschema/issues/782

terencehonles:
I bumped AppVeyor to a config which actually has Python 3.9, but I still can't reproduce the Python2 error as mentioned in https://github.com/Yelp/bravado-core/pull/383#issuecomment-946444175 (but now it's at least the same point version so it does seem like this is a Windows only issue)

PR #386, opened 2021-12-09
Use http_client from the internal spec in the `_SpecFlattener.model_discovery`.

kwist-sgr:
This will allow you to use bravado-core with the jsonschema > 4.0.

E.g. in the following case:
```python
from bravado.requests_client import RequestsClient

http_client = RequestsClient()
spec = Spec.from_dict(spec_dict, origin_url=origin_url, config=config, http_client=http_client)

# This call can throw an error without the fix
flattened_spec = spec.deref_flattened_spec
```

### DIFF
--- a/bravado_core/spec_flattening.py
+++ b/bravado_core/spec_flattening.py
@@ -318,9 +318,7 @@ class _SpecFlattener(object):
                         for uri, value in iteritems(self.known_mappings['definitions'])
                     },
                 },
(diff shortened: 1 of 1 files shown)


kwist-sgr:
And bravado-core doesn't support `jsonschema>4` at the moment - a large number of tests fail.
I guess the reason is that the last release of `bravado-core` was before `jsonschema v4` was released.

See also #385.

PR #387, opened 2021-12-15
Upstream jsonschema refresolution issue patch

rpcope1:
Schema validation with bravado-core causes "RefResolutionError" when used with newer versions of jsonschema (4.0.0+); this seems to be a problem with jsonschema itself (https://github.com/Julian/jsonschema/issues/847). This commit pins jsonschema to the last known compatible version for bravado-core until the
upstream issue is addressed.


rpcope1:
Oops, looks like this duplicates #385

PR #400, opened 2023-07-18
Remove support for end-of-life Python versions

chriskuehl:
Updating the package boilerplate a bit. There is still some additional cleanup that can be done later (e.g. removing the compat module entirely).

`make test` passes locally now, but I may need to iterate on the GitHub Actions tests for a bit before this is ready for review.


benbariteau (approved):

tox.ini
```diff
@@ -12,7 +12,7 @@ commands =
     python -m pytest --cov --capture=no --benchmark-skip {posargs:tests}
 
 [testenv:benchmark]
-basepython = python3.7
+basepython = /usr/bin/python3.10
```
macisamuele, line 15: nit. Forcing the location of the python interpreter could cause issues for developers using alternative tools to have multiple python versions installed (thinking to pyenv for example).

setup.py
```diff
@@ -47,20 +31,26 @@
...
+        "Programming Language :: Python :: 3.11",
+    ],
+    install_requires=[
+        "jsonref",
+        "jsonschema[format]>=2.5.1",
+        "python-dateutil",
+        "pyyaml",
+        'requests',
+        "simplejson",
+        "six",
```
macisamuele, line 47: nit. Possibly for a separate iteration.
If python 2 is no longer supported, six dependency would no longer be needed

PR #414, opened 2026-04-13
Add msgpack serialization support for request bodies

ltrottier-yelp:
Summary

  - Complete the half-implemented msgpack support by enabling request-side marshaling and unmarshaling (response-side already worked)
  - marshal_param now encodes body as msgpack when the operation's consumes includes only application/msgpack; defaults to JSON when both formats are present
  - unmarshal_param now inspects the request Content-Type header and decodes msgpack bodies via msgpack.loads
  - Add raw_bytes to IncomingRequest.__required_attrs__ to mirror IncomingResponse

  Breaking Changes

  - IncomingRequest subclasses must now provide a raw_bytes attribute
  - marshal_param encodes as msgpack for operations that exclusively consume application/msgpack

  Test plan

  - 16 new tests covering marshal/unmarshal with msgpack, JSON fallback, error handling, and non-body param isolation
  - Updated existing tests to supply headers on body-param mocks
  - Full test suite passes (878/878)

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -3,12 +3,14 @@ import logging
 import typing
 from functools import partial
 
+import msgpack
 import simplejson as json
 import six
 from six.moves.urllib.parse import quote
 
 from bravado_core import schema
 from bravado_core.content_type import APP_JSON
+from bravado_core.content_type import APP_MSGPACK
 from bravado_core.exception import SwaggerMappingError
 from bravado_core.marshal import marshal_schema_object
 from bravado_core.unmarshal import unmarshal_schema_object
@@ -151,12 +153,36 @@ def marshal_param(param, value, request):
         else:
             request.setdefault('data', {})[param.name] = value
     elif location == 'body':
-        request['headers']['Content-Type'] = APP_JSON
-        request['data'] = json.dumps(value)
+        if _should_use_msgpack(param):
+            request['headers']['Content-Type'] = APP_MSGPACK
+            request['data'] = msgpack.dumps(value)
+        else:
+            request['headers']['Content-Type'] = APP_JSON
+            request['data'] = json.dumps(value)
     else:
         raise SwaggerMappingError(
             "Don't know how to marshal_param with location {0}".format(location),
         )
+
+
+def _should_use_msgpack(param):
+    """Determine whether to use msgpack encoding for a body parameter.
+
+    Uses msgpack only when the operation's consumes list includes
+    application/msgpack but NOT application/json. When both are
+    present or neither is present, defaults to JSON for backward
+    compatibility.
+
+    :type param: :class:`bravado_core.param.Param`
+    :rtype: bool
+    """
+    try:
+        consumes = param.op.consumes
+    except AttributeError:
+        return False
+    if not consumes or not isinstance(consumes, (list, tuple)):
+        return False
+    return APP_MSGPACK in consumes and APP_JSON not in consumes
 
 
 def unmarshal_param(param, request):
@@ -187,19 +213,27 @@ def unmarshal_param(param, request):
         else:
             raw_value = cast_param(request.form.get(param.name, default_value))
     elif location == 'body':
-        try:
-            # TODO: verify content-type header
-            raw_value = request.json()
-        except ValueError as json_error:
-            # If the body parameter is required then we should make sure that an exception
-            # is thrown, instead if the body parameter is optional is OK-ish to assume that
-            # raw_value is the default_value
-            if param.required:
-                raise SwaggerMappingError(
-                    "Error reading request body JSON: {0}".format(str(json_error)),
-                )
-            else:
-                raw_value = default_value
+        content_type = request.headers.get('Content-Type', '').lower()
+        if content_type.startswith(APP_MSGPACK):
+            try:
+                raw_value = msgpack.loads(request.raw_bytes, raw=False)
+            except Exception as msgpack_error:
+                if param.required:
+                    raise SwaggerMappingError(
+                        "Error reading request body msgpack: {0}".format(str(msgpack_error)),
+                    )
+                else:
+                    raw_value = default_value
+        else:
+            try:
+                raw_value = request.json()
+            except ValueError as json_error:
+                if param.required:
+                    raise SwaggerMappingError(
+                        "Error reading request body JSON: {0}".format(str(json_error)),
+                    )
+                else:
+                    raw_value = default_value
     else:
         raise SwaggerMappingError(
             "Don't know how to unmarshal_param with location {0}".format(location),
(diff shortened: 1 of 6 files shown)


bravado_core/param.py
```diff
@@ -151,14 +153,38 @@ def marshal_param(param, value, request):
         else:
             request.setdefault('data', {})[param.name] = value
     elif location == 'body':
-        request['headers']['Content-Type'] = APP_JSON
-        request['data'] = json.dumps(value)
+        if _should_use_msgpack(param):
+            request['headers']['Content-Type'] = APP_MSGPACK
+            request['data'] = msgpack.dumps(value)
```
ymilki:
```suggestion
            request['data'] = msgpack.dumps(value, use_bin_type=True)
```

When marshalling, ensure bytes are encoded as msgpack bin. This ensures when unmarshalling with `raw=False`, we get the expected data types.

See  [tests/response/unmarshal_response_test.py and tests/response/validate_response_body_test.py](https://github.com/search?q=repo%3AYelp%2Fbravado-core+msgpack.dumps&type=code)

bravado_core/param.py
```diff
@@ -187,19 +213,27 @@ def unmarshal_param(param, request):
...
-                raise SwaggerMappingError(
-                    "Error reading request body JSON: {0}".format(str(json_error)),
-                )
-            else:
-                raw_value = default_value
+        content_type = request.headers.get('Content-Type', '').lower()
+        if content_type.startswith(APP_MSGPACK):
+            try:
+                raw_value = msgpack.loads(request.raw_bytes, raw=False)
+            except Exception as msgpack_error:
```
ymilki: Can you catch more explicit errors?

bravado_core/request.py
```diff
@@ -13,12 +13,13 @@ class IncomingRequest(object):
...
-        'form',     # dict of form parameters from a POST
-        'headers',  # dict of request headers
+        'path',      # dict of URL path parameters
+        'query',     # dict of parameters from the query string
+        'form',      # dict of form parameters from a POST
+        'headers',   # dict of request headers
         # TODO: may need to make this more flexible based on actual usage and/or need for a file like object # noqa
-        'files',    # dict of filename to content
+        'files',     # dict of filename to content
+        'raw_bytes',  # the bytes of the body
```
ymilki: This attempt at aligning the comments was not done correctly. Fix the alignment or just add the new entry without fixing the aligning.

tests/param/unmarshal_param_test.py
```diff
@@ -294,6 +298,130 @@ def test_body_parameter_not_present_not_required(empty_swagger_spec, body, expec
         },
     }
     param = Param(empty_swagger_spec, Mock(spec=Operation), param_spec)
-    request = Mock(spec=IncomingRequest, json=Mock(return_value=body))
+    request = Mock(spec=IncomingRequest, headers={}, json=Mock(return_value=body))
     value = unmarshal_param(param, request)
     assert expected_value == value
+
+
+def test_body_msgpack(empty_swagger_spec, param_spec):
```
ymilki, line 307: All these tests have the same pattern resulting in duplication. Can you DRY them up?

tests/param/marshal_param_test.py
```diff
@@ -355,3 +357,87 @@ def test_boolean_query_params_are_lower_case(
 def test_encode_request_param(param_type, param_value, expected_param_value):
     value = encode_request_param(param_type, 'some_name', param_value)
     assert value == expected_param_value
+
+
+def test_body_msgpack_only_consumes(empty_swagger_spec, param_spec):
```
ymilki, line 362: All these tests have the same pattern resulting in duplication. Can you DRY them up?

tests/request/IncomingRequest_test.py
```diff
@@ -26,6 +26,28 @@ class NonCompliantRequest(IncomingRequest):
...
+    class NoRawBytesRequest(IncomingRequest):
+        pass
+
+    r = NoRawBytesRequest()
+    with pytest.raises(NotImplementedError) as excinfo:
+        r.raw_bytes
+    assert 'forgot to implement' in str(excinfo.value)
+
+
+def test_raw_bytes_attr_returned():
```
ymilki: This unit test is unnecessary. It confirms `__getattr__` and `__required_attrs` during dev, but isn't necessary to keep.

tests/request/IncomingRequest_test.py
```diff
@@ -26,6 +26,28 @@ class NonCompliantRequest(IncomingRequest):
     assert 'forgot to implement' in str(excinfo.value)
 
 
+def test_missing_raw_bytes_throws_NotImplementedError():
```
ymilki: This unit test is unnecessary. It confirms `__getattr__` and `__required_attrs` during dev, but isn't necessary to keep.

ymilki (changes requested):
A couple changes and formatting requests.

bravado_core/request.py
ltrottier-yelp: In reality, there should be no alignment between lines, as a change in one line propagates unnecessary changes in other lines. The best is to not align and use double space at the end of the code (black formatter default).

bravado_core/param.py
ltrottier-yelp: Ok I will change the code to catch more explicit exceptions

tests/param/unmarshal_param_test.py
ltrottier-yelp: I removed tests that hit the same coverage and I added docstring to document why they are needed.

tests/param/marshal_param_test.py
ltrottier-yelp: I removed tests that hit the same coverage and I added docstring to document why they are needed.

bravado_core/param.py
ltrottier-yelp: I will switch to using packb instead because dumps is non canonical.

tests/param/marshal_param_test.py
ymilki: I still see a lot of new tests with the same body patterns in both test files. By DRYing them up, I mean to consolidate these multiline patterns - e.g. use fixtures, helper functions, etc - to simplify the bodies.

ltrottier-yelp: The existing tests follow the same inline-setup pattern (test_body, test_path_integer, test_header_string). They mutate param_spec directly and construct Param/Mock inline without shared fixtures or helpers. The new msgpack tests follow that same convention. Extracting helpers only for the msgpack tests would create two different styles within the same file, making the tests harder to read as a whole. Doing it selectively for the new tests would be inconsistent.

ymilki (approved):

PR #415, opened 2026-04-20
Drop python 3.7

ymilki:
## Feature

Drop Python 3.7 support, which reached end-of-life in June 2023.

## Solution

Remove the `Python :: 3.7` PyPI classifier and bump `python_requires` from `>=3.7` to `>=3.8` in `setup.py`.


sloane-shark (approved):
🚀 🚀

PR #416, opened 2026-04-20
Fix mypy and docs

ymilki:
## Problem

The docs build and mypy checks were broken due to incompatibilities with newer versions of Sphinx and stale type annotations.

## Solution

Two separate issues are addressed:

**Docs build (Sphinx 8+ compatibility)**
- Updated configs
- Pin the `docs` tox env to Python 3.10

**Mypy fixes**
- Correct the type annotation on `_unknown_type_marshaling` / `_unknown_type_unmarshaling` from `Union[Type[dict], Type[Model]]` to `Text`, matching actual usage
- Guard `_marshaling_method_primitive_type` against a `None` return from `get_type_from_schema` before passing it to `partial`
 - Removed an unused `APP_JSON` import in `tests/param/unmarshal_param_test.py` surfaced by the mypy run.

## Verification

`make docs` and `make test` pass

### DIFF
--- a/bravado_core/marshal.py
+++ b/bravado_core/marshal.py
@@ -193,7 +193,7 @@ def _no_op_marshaling(value):
 
 
 def _unknown_type_marshaling(object_type, value):
-    # type: (typing.Union[typing.Type[dict], typing.Type[Model]], typing.Any) -> NoReturn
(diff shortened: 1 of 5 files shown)


sloane-shark (approved):
🚀

PR #417, opened 2026-04-21
Use `pyproject.toml` build-system declaration

ymilki:
## Problem

The pypi github action shows the package that is built is no longer supported. This is due to mismatched assumptions with our current `setup.py/setup.cfg` with newer setuptools.

```
Checking dist/bravado_core-6.2.0.tar.gz: ERROR    InvalidDistribution: Metadata is missing required fields: Name,        
         Version.                                                               
         Make sure the distribution includes the files where those fields are   
         specified, and is using a supported Metadata-Version: 1.0, 1.1, 1.2,   
         2.0, 2.1, 2.2.                                                         
```

## Solution

Switch to more modern `pyproject.toml`

## Validation

`twine check` against sdist and wheel pass.


sloane-shark (approved):
🚀 🚀 🚀

PR #418, opened 2026-04-21
Fix pypi workflow

ymilki:
## Problem

#417 changed the build system

## Solution

Update the build instructions in the pypi workflow


sloane-shark (approved):
🚀 🚀 🚀 🚀

PR #419, opened 2026-05-04
Use standard Content-Type header for msgpack

ltrottier-yelp:
Summary

  - Replaces the internal `_should_use_msgpack()` heuristic with explicit `Content-Type` header checking for body serialization
  - Validates the requested `Content-Type` against the operation's consumes spec and raise `SwaggerMappingError` if the operation doesn't support msgpack
  - Defaults to JSON when no Content-Type is set

 Motivation

 The old `_should_use_msgpack()` function silently decided the serialization format based on the operation's consumes list. This made it impossible for callers to explicitly control request body encoding via standard HTTP headers.

With this change, the caller sets `Content-Type: application/msgpack` to opt into msgpack serialization, following standard HTTP semantics. The spec is still respected. If the operation doesn't consume msgpack, we raise an error rather than silently sending an unsupported format.

### DIFF
--- a/bravado_core/param.py
+++ b/bravado_core/param.py
@@ -153,7 +153,8 @@ def marshal_param(param, value, request):
         else:
             request.setdefault('data', {})[param.name] = value
     elif location == 'body':
-        if _should_use_msgpack(param):
+        content_type = request.get('headers', {}).get('Content-Type', '').lower()
+        if content_type == APP_MSGPACK:
             request['headers']['Content-Type'] = APP_MSGPACK
             request['data'] = msgpack.packb(value, use_bin_type=True)
         else:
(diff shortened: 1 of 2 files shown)


bravado_core/param.py
```diff
@@ -153,7 +153,8 @@ def marshal_param(param, value, request):
         else:
             request.setdefault('data', {})[param.name] = value
     elif location == 'body':
-        if _should_use_msgpack(param):
+        content_type = request.get('headers', {}).get('Content-Type', '').lower()
+        if content_type == APP_MSGPACK:
```
ymilki, line 158: The original behaviour allows msgpack according to the spec, but decides when to fallback. Here, the client decides what it wants, but we still need to follow the spec.

If the spec doesn't support msgpack, then we should not be sending msgpack. A fallback here doesn't make sense because the client won't be expecting a differently formatted response. Erroring would be more appropriate.

ymilki (changes requested):

bravado_core/param.py
ltrottier-yelp: Agree, I will change it to raise

ymilki (approved):

